PR #7, opened 2017-05-04
Adding source API's for updating source address.

vamshi-k:
***Why:***
Giving an option for the application to add the source address instead of getting resolved underneath the kernel by providing API's to set the source address

Also, added API's for updating the hostname.

@sparrc Can you please review the change ?

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -236,6 +236,17 @@ func (p *Pinger) Addr() string {
 	return p.addr
 }
 
+// SetSource sets the ip address of the source host, source should
+// be like "127.0.0.1".
+func (p *Pinger) SetSource(source string) {
+	p.source = source
+}
+
+// Addr returns the string ip address of the source host.
+func (p *Pinger) Source() string {
+	return p.source
+}
+
 // SetPrivileged sets the type of ping pinger will send.
 // false means pinger will send an "unprivileged" UDP ping.
 // true means pinger will send a "privileged" raw ICMP ping.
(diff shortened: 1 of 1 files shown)


sparrc:
why have you added a field to the struct that doesn't get used anywhere?

vamshi-k:
Can be used for debugging purposes too. There are situations where I wanted to know the hostname from which the packet stats are taken from.

sparrc:
I'd prefer if you didn't add that, because it implies that the "Hostname()" function will look up the hosts actual name somehow

vamshi-k:
@sparrc Any update on this ?

sparrc:
> I'd prefer if you didn't add that, because it implies that the "Hostname()" function will look up the hosts actual name somehow

I was waiting for you to remove the [Set]Hostname() function.

vamshi-k:
@sparrc I am have a list of hosts that I ping. List contains (hostname, ip of the host) and when I get a reply, I need to map host ip to host. Having this in a struct can give me more information about whom this pinger object refers to. If the Set/Get are to be removed, I can change the init function to add hostname as an argument (NewPinger) 

let me know if you something else that can make this work.

vamshi-k:
@sparrc how about renaming it to PingerHostname() ?

sparrc:
sorry vamshi-k but I won't be merging this. This could be easily solved in your own code by wrapping the struct and it's not the responsibility of a library to track metadata for you.

PR #10, opened 2017-06-12
remove signal handler, add Stop method

xiezhenye:
A packet should not handle signal itself. Otherwise, it will affect the user's own signal handler and cause the program can not be stopped normally and some other problem.
Maybe add a Stop method to allow user to interrupt ping.


abonec:
Totaly agree.

xiezhenye:
conflicts resolved

SuperQ:
This would be nice to get in, I'm working on building a tool that needs to be able to control the start/stop of the pings separate from the rest of the program. :smile:

PR #3, opened 2016-11-18
Check for IMCP ID

fastest963:
@sugy added a commit to check the icmp ID and verify the sequence number. This prevents other applications or pings from interfering with pinger

PR #15, opened 2017-08-22
Added ICMP Packet ID to distinguish ICMP replies in multithreading environment

webfrank:
When using the library to ping different hosts simultaneously the previous implementation, not setting/checking ICMP packet ID, will mix replies.


gm42:
Isn't this a duplicate of #3?

corny:
Our [ping package](https://github.com/digineo/go-ping) also solves this issue.

PR #18, opened 2018-01-31
Change size to Size to make it exportable

keeZey:
I have made Size exportable as i require the ability to change the
packetsize on a per host basis


dynahcatq:
I have the same issue for non-configurable packet size. LGTM and it is very useful to set packet size, @sparrc can you merge this please?

sparrc:
Thanks for the contribution! apologies for the delay getting to this

Issue #13, opened 2017-08-06
CPU utilization problem


ascotan:
There is a problem with CPU utilization on the for/select block on line 293 of ping.go.  Essentially when you have a large number of pingers running the "default" case causes the for loop to spin up the CPU.  I had 700% CPU utilization on my mac.  

The fix for this seems to be to move the default case outside of the select block.  I haven't tested all the permutations here.

```
	for {
		select {
		case <-c:
			close(p.done)
		case <-p.done:
			wg.Wait()
			return
		case <-timeout.C:
			close(p.done)
			wg.Wait()
			return
		case <-interval.C:
			err = p.sendICMP(conn)
			if err != nil {
				fmt.Println("FATAL: ", err.Error())
			}
		case r := <-recv:
			err := p.processPacket(r)
			if err != nil {
				fmt.Println("FATAL: ", err.Error())
			}
		}
                // moved outside of the select block
		if p.Count > 0 && p.PacketsRecv >= p.Count {
			close(p.done)
			wg.Wait()
			return
		}
	}
```

Pascougnou:
After some investigation I think you should also close your tickers otherwise it will eat all your CPU over time. 
cf https://github.com/SewanDevs/go-ping/commit/42df0c56848c82e922339d91aa4d8aa8ed101cf4

PR #20, opened 2018-04-15
bugfix: endless loop

xjdrew:
endless and cosume massive cpu if a ping packet lost

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -311,7 +311,6 @@ func (p *Pinger) run() {
 			if err != nil {
 				fmt.Println("FATAL: ", err.Error())
 			}
-		default:
 			if p.Count > 0 && p.PacketsRecv >= p.Count {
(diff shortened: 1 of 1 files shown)


xjdrew:
similar problem as issue #13

ping.go
```diff
@@ -311,7 +311,6 @@ func (p *Pinger) run() {
 			if err != nil {
 				fmt.Println("FATAL: ", err.Error())
 			}
-		default:
 			if p.Count > 0 && p.PacketsRecv >= p.Count {
```
sparrc: This block should be unindented and moved outside of the select clause

sparrc:
Thanks for the contribution! sorry for the delay getting to this.

sparrc:
this has now been fixed in #40

PR #30, opened 2018-08-30
fix duplicate field network and ipv4 error

gaxuhongyu:
error like:
`
vendor/github.com/sparrc/go-ping/ping.go:97:10: duplicate field name in struct literal: network
vendor/github.com/sparrc/go-ping/ping.go:98:7: duplicate field name in struct literal: ipv4
`


sparrc:
Thank you, but this was fixed by #27 already

PR #32, opened 2018-10-02
patch: compare identifier for non-root access icmp

isi-lincoln:
Commit d046b245 introduces a bug which causes ping to always fail.

This is a patch for https://github.com/sparrc/go-ping/issues/31

The source of this bug is:

```
	// Check if reply from same ID
	body := m.Body.(*icmp.Echo)
	if body.ID != p.id {
		return nil
	}
```

Which due to the selection of p.id requires that SetPrivileged is
set to true.  In the case where Privileged (i.e p.network == udp)
it is left to the kernel to set the ICMP id.

https://lwn.net/Articles/443051/  Discusses the introduction of
non-setuid-less ping.  The kernel implementation for this
interface dictates using the local port, which gets mapped into
the ping_table struct.  There is no current implementation in the
go icmp library to address this problem directly.

To address this issue, I've added a `Tracker` field for `Pinger`
as well as `IcmpData` datastructure to allow for uniquely tracking
icmp requests.  The id (as with the `id` field) is not unique,
but will statistically rare for duplicates.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -44,6 +44,7 @@
 package ping
 
 import (
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


sparrc:
Thank you for digging into this and for the contribution, much appreciated!

Issue #9, opened 2017-06-12
Multiple ping got mixed


webfrank:
I don't see any id/seq number on ping sent, running multiples ping from same host will mix results.

webfrank:
Added PR https://github.com/sparrc/go-ping/pull/15

sparrc:
this should be fixed after merging #15

PR #33, opened 2018-10-09
fix issue #9

unCleanCode:


ping.go
```diff
@@ -84,6 +84,7 @@ func NewPinger(addr string) (*Pinger, error) {
 		ipv4 = false
 	}
 
+    rand.Seed(time.Now().UnixNano())
```
SuperQ: This looks like it needs `go fmt`.

ricardoatsouza:
This also fixes this issue (https://github.com/sparrc/go-ping/issues/14).

sparrc:
I believe this issue was caused by previously the code was not checking the icmp id field, can you confirm this @ricardoatsouza ?

ricardoatsouza:
@sparrc `rand.Seed` is still necessary to guarantee that `rand.Intn(0xffff)` will not collide when running two processes at the same time.  I wrote a quick explanation here: https://github.com/sparrc/go-ping/issues/14#issuecomment-436254758

sparrc:
I'm not sure that this PR will fix the issue in all cases. Setting the random seed sets a global seed so the pingers would likely use the same seed unless one was started before the other was instantiated. I think we need to embed a random number generator with a seed into the struct.

sparrc:
see #43

ricardoatsouza:
@sparrc I see what you mean. While it does solve the problem for two running processes, it might still cause conflicts internally when using two pingers. Good point.

PR #38, opened 2018-11-04
Include adddress in Packet

SuperQ:
Include the original address in the Packet type for use in
`pinger.OnRecv()`.


sparrc:
Thanks for the contribution @SuperQ !

PR #39, opened 2018-11-04
Update to CircleCI 2.0

SuperQ:
Convert circle.yml to 2.0 version.


sparrc:
thank you for the contribution! I've migrated to Circle v2 in a separate PR so that I could get tests passing (#42), apologies for not seeing this before I started on that.

PR #49, opened 2019-02-05
Added ToS on outgoing packets

webfrank:
I added the ability to set ToS on outgoing packets.

It would be nice to get the received ToS but I think all should be ported on RAW socket to have the IP header.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -137,6 +137,9 @@ type Pinger struct {
 	// Tracker: Used to uniquely identify packet when non-priviledged
 	Tracker int64
 
+	// ToS: Type of Service for outgoing packet
+	Tos int
+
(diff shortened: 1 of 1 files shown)


webfrank:
Hi, yes as default value is zero. 

Inviato da iPhone

sparrc:
Thanks for the contribution @webfrank 

I don't see any reason why not to merge this, but can I ask why you need the ToS header? From my experience this field is very rarely used. It also seems to have gone away in IPv6.

CHTJonas:
We've discussed this internally on our Slack channel a bit and have decided to close I'm afraid. The consensus was that it's doubtful anyone actually honours the TOS packet header seriously so there's little point in merging this.

Issue #50, opened 2019-02-28
Add packet timeout callback


sajal:
I would like to be notified when a ping fails in a specified timeout.

Example I have a long running pinger hitting `8.8.8.8`, whenever a probe does not get a response within 1 second, I would like a callback to fire informing me of the same, so my code can do something with that information.

It appears that the only way to do so currently is call `Statistics()` regularly and do some math to figure out if something lossy happened recently or not.

sparrc:
@sajal This is a good idea, feel free to comment on #51 as you see fit.

chrisFang5:
@sparrc Is anyone solving this problem?

PR #51, opened 2019-03-02
fix for issue #50 (Add icmp OnTimeout callback support)

eunchong:
fix for issue #50 

Can use like this..

```Go
	pinger.OnTimeout = func(pkt *ping.Packet) {
		fmt.Printf("Request timeout for %s: icmp_seq=%d\n",
			pkt.IPAddr, pkt.Seq)
	}
```

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -62,6 +62,7 @@ const (
 	timeSliceLength  = 8
 	protocolICMP     = 1
 	protocolIPv6ICMP = 58
+	packetTimeoutSec = 5
 )
 
 var (
@@ -128,6 +129,9 @@ type Pinger struct {
 	// OnRecv is called when Pinger receives and processes a packet
 	OnRecv func(*Packet)
 
+	// OnTimeout is called when packet timeout
+	OnTimeout func(*Packet)
+
 	// OnFinish is called when Pinger exits
 	OnFinish func(*Statistics)
 
@@ -149,6 +153,8 @@ type Pinger struct {
 	id       int
 	sequence int
 	network  string
+
+	waitPackets []*Packet
 }
 
 type packet struct {
@@ -172,6 +178,12 @@ type Packet struct {
 
 	// Seq is the ICMP sequence number.
 	Seq int
+
+	// Requested time.
+	ReqTime time.Time
+
+	// if is Timeout.
+	Timeout bool
 }
 
 // Statistics represent the stats of a currently running or finished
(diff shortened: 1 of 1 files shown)


ping.go
```diff
@@ -497,6 +513,34 @@ type IcmpData struct {
...
+func (p *Pinger) checkWaitPacket() {
+	outPkt := &Packet{
+		IPAddr:  p.ipaddr,
+		Addr:    p.addr,
+		Seq:     p.sequence,
+		ReqTime: time.Now(),
+		Timeout: false,
+	}
+
+	p.waitPackets = append(p.waitPackets, outPkt)
```
sparrc: Won't this list will grow forever? Can you please implement something to keep it short?

ping.go
```diff
@@ -497,6 +513,34 @@ type IcmpData struct {
...
+
+	for _, waitPacket := range p.waitPackets {
+
+		timeDiff := time.Now().Sub(waitPacket.ReqTime)
+
+		if timeDiff.Seconds() > packetTimeoutSec && waitPacket.Timeout == false && outPkt.Nbytes <= 0 {
+			waitPacket.Timeout = true
+
+			timeoutHandler := p.OnTimeout
```
sparrc: this assignment isn't necessary

ping.go
```diff
@@ -497,6 +513,34 @@ type IcmpData struct {
 	Tracker int64
 }
 
+func (p *Pinger) checkWaitPacket() {
```
sparrc: this function should return immediately if `p.OnTimeout == nil`

CHTJonas:
Hi @eunchong - been a long time but are you interested in bringing this up-to-date and addressing the feedback? If not we'll close the PR.

CHTJonas:
Poke @eunchong 😉

CHTJonas:
Closing due to inactivity.

yxxhero:
@CHTJonas Can it asign to me?

CHTJonas:
IMO the solution to this PR, #165, #50 and #163 are the same and involve adding extra data/state to the `map[uuid.UUID]map[int]struct{}{}` type that we're using to keep track of packets locally. That anonymous `struct{}{}` would probably need to become something more well-defined.

IMO we don't want (too much) state in ICMP messages/headers if we can possibly help it. Trouble is that this doesn't yet exist in `master` yet so I think all of this is kinda blocked on #152. I think it would be best for you to jump on the #go-ping channel of the [Gophers Slack](https://invite.slack.golangbridge.org/) and discuss with us.

PR #55, opened 2019-05-21
Expose TTL from received packets

mchurichi:
This is @clinta's work (https://github.com/sparrc/go-ping/pull/19) rebased.

### DIFF
--- a/cmd/ping/ping.go
+++ b/cmd/ping/ping.go
@@ -65,8 +65,8 @@ func main() {
 	}()
 
 	pinger.OnRecv = func(pkt *ping.Packet) {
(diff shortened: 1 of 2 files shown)


mchurichi:
@sparrc can we have this merged?

sparrc:
Thanks for rebasing the work @mchurichi, but the proper way to do it would be to cherry-pick his actual commit, so he will still get credit for it.

mchurichi:
@sparrc sure thing!, it was never my intention to stole @clinta's work, just wanted to expedite this... here is the updated branch with the original commit, thanks!

PR #59, opened 2019-05-30
Make `processPacket` more performant

### LINKED ISSUE #44: Use gob over json encoding
json is very slow, comparing normal C `/bin/ping` to Go `ping`, there's a lot of added latency.

```console
$ /bin/ping -c 10 localhost
PING localhost (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.085 ms
64 bytes from localhost (127.0.0.1): icmp_seq=2 ttl=64 time=0.071 ms
64 bytes from localhost (127.0.0.1): icmp_seq=3 ttl=64 time=0.076 ms
64 bytes from localhost (127.0.0.1): icmp_seq=4 ttl=64 time=0.081 ms
64 bytes from localhost (127.0.0.1): icmp_seq=5 ttl=64 time=0.063 ms
64 bytes from localhost (127.0.0.1): icmp_seq=6 ttl=64 time=0.100 ms
64 bytes from localhost (127.0.0.1): icmp_seq=7 ttl=64 time=0.082 ms
64 bytes from localhost (127.0.0.1): icmp_seq=8 ttl=64 time=0.050 ms
64 bytes from localhost (127.0.0.1): icmp_seq=9 ttl=64 time=0.101 ms
64 bytes from localhost (127.0.0.1): icmp_seq=10 ttl=64 time=0.102 ms
```

```console
$ ./ping --privileged -c 10 localhost
PING localhost (127.0.0.1):
62 bytes from 127.0.0.1: icmp_seq=0 time=638.779µs
62 bytes from 127.0.0.1: icmp_seq=1 time=188.839µs
62 bytes from 127.0.0.1: icmp_seq=2 time=148.038µs
62 bytes from 127.0.0.1: icmp_seq=3 time=473.847µs
62 bytes from 127.0.0.1: icmp_seq=4 time=336.945µs
62 bytes from 127.0.0.1: icmp_seq=5 time=344.765µs
62 bytes from 127.0.0.1: icmp_seq=6 time=270.598µs
62 bytes from 127.0.0.1: icmp_seq=7 time=96.763µs
62 bytes from 127.0.0.1: icmp_seq=8 time=415.721µs
62 bytes from 127.0.0.1: icmp_seq=9 time=195.21µs
```

glinton:
Currently, encoding with `json`, the `processPacket` function is ~7x slower than using straight binary encoding.

```
# json encoding
BenchmarkProcessPacketOld-8   	 1000000	      2131 ns/op
# binary encoding
BenchmarkProcessPacket-8      	 5000000	       283 ns/op
```

I feel like this resolves #44 even though it isn't using gob, as it seems the goal of 44 was to use a faster encoding method.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -46,7 +46,6 @@ package ping
 import (
 	"bytes"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"math"
 	"math/rand"
@@ -486,7 +485,8 @@ func (p *Pinger) processPacket(recv *packet) error {
 		}
 
 		if len(pkt.Data) < timeSliceLength+trackerLength {
-			return errors.New("insufficient data received")
+			fmt.Printf("insufficient data received; got: %d %v",
+				len(pkt.Data), pkt.Data)
 		}
 
 		tracker := bytesToInt(pkt.Data[timeSliceLength:])
--- a/ping.go
+++ b/ping.go
@@ -44,7 +44,9 @@
 package ping
 
 import (
-	"encoding/json"
+	"bytes"
+	"encoding/binary"
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
@@ -60,6 +62,7 @@ import (
 
 const (
 	timeSliceLength  = 8
+	trackerLength    = 8
 	protocolICMP     = 1
 	protocolIPv6ICMP = 58
 )
@@ -438,6 +441,7 @@ func (p *Pinger) recvICMP(
 }
 
 func (p *Pinger) processPacket(recv *packet) error {
+	receivedAt := time.Now()
 	var bytes []byte
 	var proto int
 	if p.ipv4 {
@@ -455,32 +459,12 @@ func (p *Pinger) processPacket(recv *packet) error {
 	var m *icmp.Message
 	var err error
 	if m, err = icmp.ParseMessage(proto, bytes[:recv.nbytes]); err != nil {
-		return fmt.Errorf("Error parsing icmp message")
+		return fmt.Errorf("error parsing icmp message: %s", err.Error())
 	}
 
 	if m.Type != ipv4.ICMPTypeEchoReply && m.Type != ipv6.ICMPTypeEchoReply {
 		// Not an echo reply, ignore it
 		return nil
-	}
-
-	body := m.Body.(*icmp.Echo)
-	// If we are priviledged, we can match icmp.ID
-	if p.network == "ip" {
-		// Check if reply from same ID
-		if body.ID != p.id {
-			return nil
-		}
-	} else {
-		// If we are not priviledged, we cannot set ID - require kernel ping_table map
-		// need to use contents to identify packet
-		data := IcmpData{}
-		err := json.Unmarshal(body.Data, &data)
-		if err != nil {
-			return err
-		}
-		if data.Tracker != p.Tracker {
-			return nil
-		}
 	}
 
 	outPkt := &Packet{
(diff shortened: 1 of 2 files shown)


ping.go
```diff
@@ -533,42 +526,41 @@ func (p *Pinger) sendICMP(conn *icmp.PacketConn) error {
...
+		if _, err := conn.WriteTo(msgBytes, dst); err != nil {
 			if neterr, ok := err.(*net.OpError); ok {
 				if neterr.Err == syscall.ENOBUFS {
 					continue
 				}
 			}
 		}
-		p.PacketsSent += 1
-		p.sequence += 1
+		p.PacketsSent++
```
sparrc: 👍

ping.go
```diff
@@ -492,18 +476,32 @@ func (p *Pinger) processPacket(recv *packet) error {
...
+		// If we are priviledged, we can match icmp.ID
+		if p.network == "ip" {
+			// Check if reply from same ID
+			if pkt.ID != p.id {
+				return nil
+			}
+		}
+
+		if len(pkt.Data) < timeSliceLength+trackerLength {
+			return errors.New("insufficient data received")
```
sparrc: could you change this to `fmt.Errorf` to avoid adding in the errors package?

sparrc: please also add some details about the packet that might be useful for debugging

glinton: For future reference, are you concerned about the binary size or just a larger `import` section?
As an aside, technically `errors` is already included by [`fmt`](https://golang.org/src/fmt/print.go) so there is no binary growth by importing it.

sparrc: ah, I didn't realize fmt was already using errors, thx for the heads-up

sparrc:
@glinton can you rebase your branch please?

glinton:
I don't generally like rebasing and force pushing a single commit to a PR as it doesn't preserve review comments. I also get nervous as it seems like it begs for another review as I could have added garbage code before pushing.
So, if I did it wrong, I apologize, I can force push the other commits back up and rebase again in the manner you prefer.

sparrc:
> I don't generally like rebasing and force pushing a single commit to a PR as it doesn't preserve review comments. I also get nervous as it seems like it begs for another review as I could have added garbage code before pushing.
> So, if I did it wrong, I apologize, I can force push the other commits back up and rebase again in the manner you prefer.

Sorry I should have specified, merging in master would have worked as well, the UI was not letting me merge so I think the previous merge commit may have missed something.

ping_test.go
```diff
@@ -264,3 +271,174 @@ func AssertFalse(t *testing.T, b bool) {
...
+		nbytes: len(msgBytes),
+		bytes:  msgBytes,
+	}
+
+	processPacket := func(p *Pinger, recv *packet) error {
+		var bytes []byte
+		var proto int
+		if p.ipv4 {
+			if p.network == "ip" {
+				bytes = ipv4Payload(recv.bytes)
```
sparrc: This should be just `recv` now, I will add unit testing and `go vet` to the circleci config to catch this in the future

ping.go
```diff
@@ -492,18 +475,33 @@ func (p *Pinger) processPacket(recv *packet) error {
...
+		// If we are priviledged, we can match icmp.ID
+		if p.network == "ip" {
+			// Check if reply from same ID
+			if pkt.ID != p.id {
+				return nil
+			}
+		}
+
+		if len(pkt.Data) < timeSliceLength+trackerLength {
+			fmt.Printf("insufficient data received; got: %d %v",
```
sparrc: This should be `return fmt.Errorf(...`

ping_test.go
```diff
@@ -264,3 +271,174 @@ func AssertFalse(t *testing.T, b bool) {
...
+		bytes:  msgBytes,
+		ttl:    24,
+	}
+
+	for k := 0; k < b.N; k++ {
+		pinger.processPacket(&pkt)
+	}
+}
+
+func BenchmarkProcessPacketOld(b *testing.B) {
```
sparrc: You can just remove the "old" benchmark going forward

PR #62, opened 2019-06-05
Don't separate ping response header from received bytes

sparrc:
It's not clear to me that golang's ICMP parsing library actually requires separating the "payload" from the header of the response message. This removes ipv4Payload function entirely. Needs thorough testing before merging.

closes https://github.com/sparrc/go-ping/issues/61

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -441,23 +441,16 @@ func (p *Pinger) recvICMP(
 
 func (p *Pinger) processPacket(recv *packet) error {
 	receivedAt := time.Now()
-	var bytes []byte
 	var proto int
 	if p.ipv4 {
-		if p.network == "ip" {
-			bytes = ipv4Payload(recv)
-		} else {
-			bytes = recv.bytes
-		}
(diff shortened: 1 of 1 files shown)


sparrc:
cc @AlirieGray @glinton 

As you were both in this code recently, can you please review?

sparrc:
also cc @SuperQ

SuperQ:
I guess I don't know the Go ICMP stuff well enough to give a good review. But, if it works, it works. :grin:

glinton:
> It's not clear to me that golang's ICMP parsing library actually requires separating the "payload" from the header of the response message.

From what I can tell, it isn't required as it's handled in the [`ReadFrom` function](https://github.com/golang/net/blob/c98bf442cc908065c4df3decd0c6a9523c7896fa/ipv4/payload_cmsg.go#L40-L47)

chn-lee-yumi:
Hello, this really fixed the bug, but I can't get this code using `go get`. Had it merged to master branch?

PR #65, opened 2019-06-06
Don't resolve when creating Pinger

SuperQ:
Move the DNS resolver out of the NewPinger() function in order to allow
adjusting of IPv4 vs IPv6 DNS resolution before running. This also
allows the user to verify resolution.
* Create new `New()` method that returns a bare default struct.
* Create new `Resolve()` method.
* Call `Resolve()` from `SetAddr()`.
* Call `Resolve()` automatically from `Run()`.

Update ping command for simplifed return values of `NewPinger()`.

Fixes: https://github.com/sparrc/go-ping/issues/64

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -67,41 +67,37 @@ const (
 )
 
 var (
-	ipv4Proto = map[string]string{"ip": "ip4:icmp", "udp": "udp4"}
-	ipv6Proto = map[string]string{"ip": "ip6:ipv6-icmp", "udp": "udp6"}
+	ipv4Proto = map[string]string{"icmp": "ip4:icmp", "udp": "udp4"}
+	ipv6Proto = map[string]string{"icmp": "ip6:ipv6-icmp", "udp": "udp6"}
 )
 
-// NewPinger returns a new Pinger struct pointer
-func NewPinger(addr string) (*Pinger, error) {
-	ipaddr, err := net.ResolveIPAddr("ip", addr)
-	if err != nil {
-		return nil, err
-	}
-
-	var ipv4 bool
-	if isIPv4(ipaddr.IP) {
-		ipv4 = true
-	} else if isIPv6(ipaddr.IP) {
-		ipv4 = false
-	}
-
+// New returns a new Pinger struct pointer.
+func New(addr string) *Pinger {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	return &Pinger{
-		ipaddr:   ipaddr,
-		addr:     addr,
-		Interval: time.Second,
-		Timeout:  time.Second * 100000,
 		Count:    -1,
-		id:       r.Intn(math.MaxInt16),
-		network:  "udp",
-		ipv4:     ipv4,
+		Interval: time.Second,
 		Size:     timeSliceLength,
+		Timeout:  time.Second * 100000,
 		Tracker:  r.Int63n(math.MaxInt64),
+
+		addr:     addr,
 		done:     make(chan bool),
-	}, nil
+		id:       r.Intn(math.MaxInt16),
+		ipaddr:   nil,
+		ipv4:     false,
+		network:  "ip",
+		protocol: "udp",
+	}
 }
 
-// Pinger represents ICMP packet sender/receiver
+// NewPinger returns a new Pinger and resolves the address.
+func NewPinger(addr string) (*Pinger, error) {
+	p := New(addr)
+	return p, p.Resolve()
+}
+
+// Pinger represents a packet sender/receiver.
 type Pinger struct {
 	// Interval is the wait time between each packet send. Default is 1s.
 	Interval time.Duration
@@ -152,7 +148,10 @@ type Pinger struct {
 	size     int
 	id       int
 	sequence int
-	network  string
+	// network is one of "ip", "ip4", or "ip6".
+	network string
+	// protocol is "icmp" or "udp".
+	protocol string
 }
 
 type packet struct {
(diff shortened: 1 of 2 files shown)


ping.go
```diff
@@ -101,6 +107,26 @@ func NewPinger(addr string) (*Pinger, error) {
...
 
+// PingerConf represents options to configuring a pinger.
+type PingerConf struct {
+	// The address to ping.
+	Address string
+	// The network type from net.ResolveIPAddr(). ("ip", "ip4", "ip6")
+	Network string
+}
+
+func NewPingerConf(n string, addr string) PingerConf {
```
sparrc: I would prefer if the arguments to this function only included required parameters.

ping.go
```diff
@@ -101,6 +107,26 @@ func NewPinger(addr string) (*Pinger, error) {
 	}, nil
 }
 
+// PingerConf represents options to configuring a pinger.
+type PingerConf struct {
+	// The address to ping.
```
sparrc: Please indicate in the comments which parameters are required and which are optional.

ping.go
```diff
@@ -71,9 +71,15 @@ var (
...
 
-// NewPinger returns a new Pinger struct pointer
+// NewPinger returns a new Pinger struct pointer based on an address string.
 func NewPinger(addr string) (*Pinger, error) {
-	ipaddr, err := net.ResolveIPAddr("ip", addr)
+	conf := NewPingerConf("ip", addr)
+	return NewPingerWithConf(conf)
+}
+
+// NewPignerWithConf returns a new Pinger struct poitner based on a PingerConf.
```
sparrc: spelling

sparrc:
@SuperQ I agree that we should have a way to manually choose between ipv6 and ipv4, but now that I review this PR I'm thinking that it makes the package API a bit confusing.

the `Pinger` struct itself contains most of the config already....so it's strange to have a separate `PingerConf` struct that actually only controls the configuration variables that get evaluated at the time Pinger is instantiated.

Maybe your original idea would be more correct, but implemented as another function like `NewPingerWithNetwork(network string, addr string)`

SuperQ:
Yea, the issue is that NewPinger() does DNS resolution.

What if we moved the `net.ResolveIPAddr()` out of `NewPinger()`, and made it part of `Run()`? Or did it more automatically if no IP is set? This way we can skip doing a `net.ResolveIPAddr()` if the user makes the address an IP.

SuperQ:
Ok, I complete re-did this. I moved the DNS resolution out of `NewPinger()`. This makes it easier to finish setting up the Pinger. It would probably be good to create a release tag in GitHub before merging this so people can lock the dependency due to the minor breaking change that `NewPinger()` doesn't return an error anymore.

sparrc:
Sorry @SuperQ, I like the behavior you've implemented but I don't think it's worth making a breaking change over this. Could you please implement this as a separate NewPinger function?

SuperQ:
I think it's a bit excessive to create a whole separate function just to avoid a breaking change in a library that's not 1.0, and doesn't have any tagged releases at all.

I could make `NewPinger()` return the same, but will always return nil. But again, I think it's a better to just break things and move forward with better releases than try and keep things the same forever.

SuperQ:
I reworked this slightly. Now it has a simple `New()` method, and `NewPinger()` now has the original behavior of creating and resolving in one go.

SuperQ:
@sparrc Does this look better now?

sparrc:
sorry for the delay in reviewing, I recently had a baby so have been a bit tied up, will get to it soon

SuperQ:
No worries. Congrats! :tada:

ping.go
```diff
@@ -442,7 +486,7 @@ func (p *Pinger) recvICMP(
 func (p *Pinger) processPacket(recv *packet) error {
 	receivedAt := time.Now()
 	var proto int
-	if p.ipv4 {
+	if *p.ipv4 {
```
sparrc: Is there any chance any of these if-statements could be reached with p.ipv4 still being unset? Just for the sake of safety could you change these to:

```
if p.ipv4 == nil || *p.ipv4 {
    ...
}
```

ping.go
```diff
@@ -225,27 +226,47 @@ func (p *Pinger) SetIPAddr(ipaddr *net.IPAddr) {
...
 	p.addr = ipaddr.String()
-	p.ipv4 = ipv4
 }
 
 // IPAddr returns the ip address of the target host.
 func (p *Pinger) IPAddr() *net.IPAddr {
 	return p.ipaddr
 }
 
+// Rsolve resolves the Pinger address and sets network type.
```
sparrc: spelling (Rsolve)

sparrc (dismissed):

ping.go
SuperQ: We have to pass through a `Resolve()` cleanly which will always set the packet ipv4.

If `p.ipv4` were nil at this point, I would probably throw an error or `panic()`.

sparrc: it will panic already when you try to dereference a nil pointer. That's what I want to protect against, I don't want it to panic because libraries generally never should.

SuperQ: But it's completely invalid to continue here if `p.ipv4 == nil`, which is what your change is suggesting.

SuperQ: I was thinking about another way to solve this concern. Rather than depend on the set/unset state of the ipv4, we can have a second `resolved` bool in the Pinger struct.

sparrc: you could also make it an enum, something like:
```go
type networkType int

const (
	unset networkType = iota
	ipv4
	ipv6
)
```
I realize it's not valid for it to be unset, but the current behavior is for there to be a defacto 'default' behavior and NOT to panic. This library should avoid panics at all costs.

SuperQ: Hrm, that's an interesting idea. I'll think about the options and spend some time on it soon. I'm starting to deploy my smokeping tool to production and it would be good to be able to better control protocol selection. :smile:

SuperQ:
Ok, I cleaned up the problem with the nil pointer panic. I realized that we can simply key on `ipaddr` being nil to determine if the resolve has happened.

SuperQ:
Ping, what do you think of the new approach?

SuperQ:
Any chance to review the new version?

SuperQ:
Any chance you can look at this again?

CHTJonas (approved):
Looks good to me and happy with the approach this takes RE: resolving hostnames, picking the IP protocol to use and maintaining API compatibility. Nit pick: could do with `go fmt` being run.

SuperQ:
Fixed `go fmt`. I'll add that to the tests.

PR #66, opened 2019-06-06
Provide a slice with per-packet TTLs in the statistics

rogerpueyo:
A couple of commits:
 - Fix linitng for TTL: rename Ttl to TTL so that the linter does not complain
 - Provide TTLs in the stats: the real thing, to add a slice in Statistics() with the TTLs

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -127,6 +127,9 @@ type Pinger struct {
 	// rtts is all of the Rtts
 	rtts []time.Duration
 
+	// ttls is all of the Ttls
+	ttls []int
+
 	// OnRecv is called when Pinger receives and processes a packet
 	OnRecv func(*Packet)
 
@@ -215,6 +218,9 @@ type Statistics struct {
 	// StdDevRtt is the standard deviation of the round-trip times sent via
 	// this pinger.
 	StdDevRtt time.Duration
+
+	// Ttts is all of the TTLs sent via this pinger.
+	Ttls []int
 }
 
 // SetIPAddr sets the ip address of the target host.
@@ -381,6 +387,7 @@ func (p *Pinger) Statistics() *Statistics {
 		IPAddr:      p.ipaddr,
 		MaxRtt:      max,
 		MinRtt:      min,
+		Ttls:        p.ttls,
 	}
 	if len(p.rtts) > 0 {
 		s.AvgRtt = total / time.Duration(len(p.rtts))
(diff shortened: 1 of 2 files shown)


SuperQ:
Hrm, these growing slices are a bit problematic for my use case. I have a process to run pingers for months at a time. I process the output of each response as they come in.

sparrc:
I guess that the slice of RTTs would have the same issue then @SuperQ ? Perhaps we need a configurable way to limit the growth of these slices.

SuperQ:
@sparrc Yes, maybe a `KeepHistory bool` on the Pinger config.

ping.go
```diff
@@ -215,6 +218,9 @@ type Statistics struct {
 	// StdDevRtt is the standard deviation of the round-trip times sent via
 	// this pinger.
 	StdDevRtt time.Duration
+
+	// Ttts is all of the TTLs sent via this pinger.
+	Ttls []int
```
sparrc: For consistency this should probably be TTLs

ping.go
```diff
@@ -179,7 +182,7 @@ type Packet struct {
 	Seq int
 
 	// TTL is the Time To Live on the packet.
-	Ttl int
+	TTL int
```
sparrc: I know it's only recently been merged, but this is now a breaking change and I'm not sure it's worth it just for the sake of linting.

sparrc: Not saying it's correct, but as Ttl it's also consistent with Rtt.

sparrc:
> @sparrc Yes, maybe a KeepHistory bool on the Pinger config.

I was thinking a configurable length, like

```go
// KeepHistory controls how many RTT and TTL values to keep for the Statistics struct
// This probably only needs to be modified to control memory consumption on long-running pingers
KeepHistory int
```

That way long-running pingers could still have a history of RTTs and TTLs

CHTJonas:
Sorry, but I'm afraid I'm not a huge fan of this change. I get where you're coming from @rogerpueyo but IMO keeping track of all the RTTs is something that should be done by the end application and not the library. The application developer will likely be aware of what kind of performance is acceptable so I think it's fair to punt keeping track of this off to them, whether they choose an expanding slice or something else.

SuperQ:
@CHTJonas I agree. I there are things we can do in the library to make this easier, we should. But the library itself shouldn't handle this.

rogerpueyo:
OK, closing PR then.

PR #70, opened 2019-07-23
fix deadlock when sending interval is very short

huwenbiao:
When Interval is very small, deadlock may happen:
1. run() wait for `wg.wait()`;
2. recvICMP() wait for `recv <- &packet{bytes: bytes, nbytes: n, ttl: ttl}`.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -313,11 +313,11 @@ func (p *Pinger) run() {
 	for {
 		select {
 		case <-p.done:
-			wg.Wait()
+			safeWait(&wg, recv)
 			return
 		case <-timeout.C:
 			close(p.done)
-			wg.Wait()
+			safeWait(&wg, recv)
 			return
 		case <-interval.C:
 			if p.Count > 0 && p.PacketsSent >= p.Count {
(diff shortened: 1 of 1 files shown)


9er:
Awesome! Thanks for posting this! I just hit this bug today and couldn't find the race-condition in my code, so I searched Github and found this.

I tried the patch and it fixed the problem for me.

adrianosela:
can we get this in please? @sparrc

ping.go
```diff
@@ -335,12 +335,19 @@ func (p *Pinger) run() {
...
-			wg.Wait()
+			safeWait(&wg, recv)
 			return
 		}
 	}
 }
 
+func safeWait(wg *sync.WaitGroup, ch chan *packet) {
+	if len(ch) > 0 {
+		<-ch
```
sparrc: why not process the packet instead of just dropping it? ie, instead of `<-ch` could we have this?:

```go
r := <-ch
p.processPacket(r)
```

sparrc:
Sorry for the delay reviewing this. Would like to understand the wider implication of simply dropping packets here.

ping.go
huwenbiao: Yes, it's better to process it.

CHTJonas:
Hello, long time no activity! Please could you rebase this on top of the latest `master` and provide an example to reproduce the bug? Thank you.

SuperQ:
I'm wondering if this is obsolete with all of the recent changes.

PR #72, opened 2019-07-29
change async Stop to golang context style

chennqqi:


cmd/ping/ping.go
```diff
@@ -1,13 +1,14 @@
...
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
 	"os/signal"
 	"time"
 
-	"github.com/sparrc/go-ping"
+	"github.com/chennqqi/go-ping"
```
SuperQ: This should stay `sparrc`

Darigaaz:
Bad idea.
Its anti-pattern to use `context` this way.
`context` does not provide acknowledgement and it wont be "fixed" until at least go 2.

SuperQ:
I also highly object to introducing GOTO label jumps in the code.

ping.go
```diff
@@ -397,13 +407,14 @@ func (p *Pinger) Statistics() *Statistics {
 func (p *Pinger) recvICMP(
 	conn *icmp.PacketConn,
 	recv chan<- *packet,
-	wg *sync.WaitGroup,
+	ctx context.Context,
 ) {
-	defer wg.Done()
+
+FOR_LOOP:
```
SuperQ: Let's not introduce GOTO label jumps, this is not a good pattern to use.

CHTJonas:
Closing. @chennqqi if you want to rebase this to bring it up to date and then address the feedback regarding the use of GOTO then we can maybe discuss again and perhaps reopen, but I don't think there is much support for using `context` in this way. Thanks for the PR though.

PR #79, opened 2020-01-17
Fixed lockup in recvICMP code.

asmpro:
I noticed lockup in

recv <- &packet{bytes: bytes, nbytes: n, ttl: ttl}:

when there are multiple CPU cores available and there are multiple pings running in parallel.


CHTJonas:
Dupe of #85

PR #80, opened 2020-02-10
fix: prevent panic when close channel

oilbeater:
We meet some panics about close a closed channel. The done channel should only be closed once.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -144,6 +144,7 @@ type Pinger struct {
 
 	// stop chan bool
 	done chan bool
+	once sync.Once
 
 	ipaddr *net.IPAddr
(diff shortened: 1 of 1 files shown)


CHTJonas:
@oilbeater, please could you rebase and provide a sample of code to replicate the issue (even if the issue only occurs intermittently). Thanks!

CHTJonas:
Going over this a second time I can see that you've also cherry-picked 325a517 into your branch; in general we prefer that each branch/PR only addresses a single issue at a time. Once that commit is removed this PR reduces down to a duplicate of #83 so I shall close in favour of that instead. Thanks for the contribution though!

Issue #16, opened 2017-11-03
Library should return errors and not printf messages


Springuin:
It's probably just my opinion, but I think having Printf's when an error occurs is not the best option. There is no easy way to divert the error messages to something else, so you practically have no choice but to have the error messages end up at stdout.
May I suggest adding a feature to divert the output (to a different stream, to a logger, ....) or adding a function that disables the messages and just return an error?

CHTJonas:
Thanks for raising! We're aware of this and agree that it needs fixing; we're looking into how best to go about this.

stevenaldinger:
the PR to close this issue looks like it's covering the logging case but not the error case. I think that's an awesome start but wondering if you were saying you agree that it needs to be passing back errors too or just that there needs to be a way to control log output?

CHTJonas:
In an ideal world, I think the library should definitely be returning errors and not logging/printing anything. Recently, #81 made the necessary changes to return from `pinger.Run()` most of the errors that can occur.

There are some parts of the library where an error can be raised inside a goroutine; we've got a few ideas how to handle this such as an integer error counter, channel or maybe even a [tomb](https://blog.labix.org/2011/10/09/death-of-goroutines-under-control) but we'd welcome any other options/opinions - PRs being even better!

SuperQ:
Another thing we could do is plumb in a logger, like [we do here](https://pkg.go.dev/github.com/prometheus/client_golang@v1.10.0/prometheus/promhttp#HandlerOpts)

lrstanley:
I'd figure I'd make a comment even though the logger interface has already been implemented... Though, if trying to keep from breaking changes I'm not sure it matters much.

I don't think there is any case for this library having such a feature/implementation. Generally speaking (& especially with libraries like this), one shouldn't do any logging at all in the package. They should return errors that the user of the library can handle if they wish to do so (and log in the way they see fit).

1. Logging is a rather opinionated thing, with many different implementations. When this library logs information, it's useless information since the end user can't do much with it. What failed? Why? Any metadata information? How do you propagate that up? etc. This is why the standard library doesn't have a `log.Logger` interface yet.
2. Returning the errors (that include information about the failure), users can decide to log, track metrics, etc. Can't easily do that with a logger by itself (esp. if it's not a structured logger -- but supporting just a structured logger means those who don't have one, are also at a loss, thus another reason why it doesn't make much sense).
3. Generally logging should be done by the destination author only, with **few exceptions**.

[...]

PR #81, opened 2020-03-05
Return error from Run()

maddie:
The current situation is the error messages are being printed with `fmt.Printf` or `fmt.Println`. In some cases, the caller might want to handle these errors by themselves without printing them.

This PR fixes the issue.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -295,9 +295,8 @@ func (p *Pinger) Privileged() bool {
 // Run runs the pinger. This is a blocking function that will exit when it's
 // done. If Count or Interval are not specified, it will run continuously until
 // it is interrupted.
-func (p *Pinger) Run() {
+func (p *Pinger) Run() (err error) {
 	var conn *icmp.PacketConn
-	var err error
 	if p.ipaddr == nil {
 		err = p.Resolve()
 	}
@@ -305,17 +304,17 @@ func (p *Pinger) Run() {
 		return
 	}
 	if p.ipv4 {
-		if conn = p.listen(ipv4Proto[p.protocol]); conn == nil {
+		if conn, err = p.listen(ipv4Proto[p.protocol]); err != nil {
 			return
 		}
-		if conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true) != nil {
+		if err = conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true); err != nil {
 			return
 		}
 	} else {
-		if conn = p.listen(ipv6Proto[p.protocol]); conn == nil {
+		if conn, err = p.listen(ipv6Proto[p.protocol]); err != nil {
 			return
 		}
-		if conn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true) != nil {
+		if err = conn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true); err != nil {
 			return
 		}
 	}
@@ -330,7 +329,7 @@ func (p *Pinger) Run() {
 
 	err = p.sendICMP(conn)
 	if err != nil {
-		fmt.Println(err.Error())
+		return err
 	}
 
 	timeout := time.NewTicker(p.Timeout)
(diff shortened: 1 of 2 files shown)


SuperQ:
Would you please rebase this change?

maddie:
@SuperQ can you help to see if the error should be returned in the following lines?

https://github.com/go-ping/ping/blob/54d38cd396bf9c3aa05cefb7f3983047a2ad0dbf/ping.go#L356
https://github.com/go-ping/ping/blob/54d38cd396bf9c3aa05cefb7f3983047a2ad0dbf/ping.go#L361

In the old code, the channel is closed and returns, but now it seems to continue if error occurs.

SuperQ:
Good question, those two lines are a bit weird. They're "fatal", but continue anyway. I would say for now, no, they should stay as print statements to maintain the current error catching behavior.

maddie:
It would seem weird to print these lines when users are using this library, since they might want to handle the errors themselves and don't want anything else printed to the stdout.

I'll leave them as-is with a `FIXME` note above, and review it later.

SuperQ:
Agreed, it's something that needs fixing.

CHTJonas (approved):
LGTM! 👍

ping.go
```diff
@@ -295,27 +295,26 @@ func (p *Pinger) Privileged() bool {
 // Run runs the pinger. This is a blocking function that will exit when it's
 // done. If Count or Interval are not specified, it will run continuously until
 // it is interrupted.
-func (p *Pinger) Run() {
+func (p *Pinger) Run() (err error) {
```
SuperQ, line 298: Hrm, I'm not sure what is the best Go style here. We're sometimes using named, and sometimes using un-named returns.

SuperQ (commented):
Adding an error to recvICMP() produces this linter failure:

```
$ golangci-lint run
ping.go:328:15: Error return value of `p.recvICMP` is not checked (errcheck)
	go p.recvICMP(conn, recv, &wg)
```

SuperQ:
Related to the FIXME, https://github.com/go-ping/ping/issues/16 should be good to track this.

ping.go
CHTJonas: I was wondering about this as well. Is there a more idiomatic go way?

maddie: IIRC both are the idiomatic way. I'm using named return in this case so less code needs to be changed because the original code declares a `var err error` in the scope.

maddie:
> Adding an error to recvICMP() produces this linter failure:
> 
> ```
> $ golangci-lint run
> ping.go:328:15: Error return value of `p.recvICMP` is not checked (errcheck)
> 	go p.recvICMP(conn, recv, &wg)
> ```

I can't think of a good way of handling this error, I think this problem falls in line with the `FIXME`s. It would be best to allow users to set a `logger` so we can just output the errors to the logger.

maddie:
I've added a `logger` interface, this should fix the problems.

SuperQ:
For the FIXMEs, I had a slight different idea. I was going to add error counters and a switch to allow the user to pick if they wanted to return on error or not.

SuperQ:
Can we split out the logger function change into a separate MR? I'd like to fix one thing at a time.

ping.go
SuperQ: From https://tour.golang.org/basics/7

> Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions

I think `Run()` qualifies as long enough that we should use explicit `return err`, rather than naked `return`s.

maddie:
Moved the logger commit away from this PR. Linter now complains the `p.recvICMP` error not being handled.

Also changed `Run()` to explicit return.

ping.go
```diff
@@ -330,7 +330,7 @@ func (p *Pinger) Run() {
 
 	err = p.sendICMP(conn)
```
SuperQ, line 331: Adding a nolint here should fix the tests for now.

```suggestion
	err = p.sendICMP(conn) //nolint:errcheck
```

SuperQ:
Adding `//nolint:errcheck` to ping.go:329 should silence the linter issue.

SuperQ (approved):
LGTM

maddie:
Ahh, sorry, I pushed to the wrong branch, I'll revert soon.

CHTJonas (approved):
LGTM 👍🚀💯

PR #86, opened 2020-03-31
add batch ping

vearne:


CHTJonas:
There's a myriad of issues with this and at this point it's very stale. I'm going to close for now but if you want to jump on our Slack channel we can discuss including this in a fresh PR.

PR #89, opened 2020-07-21
Bugfix: NewPinger shoud fail when ip is empty.

hustclf:
When NewPinger with empty ip addr,  ping.Run will hangs forever.

for example
```
func main()  {
	pinger, err := NewPinger("")
	pinger.Run()
}
```

this pr check if addr is empty at first, and return error when it is empty.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -46,6 +46,7 @@ package ping
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"math"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


CHTJonas (changes requested):
Hi @hustclf, thanks for this. Could you please rebase and move the error checking/returning into the `Resolve()` function? Thanks!

hustclf:
@CHTJonas  please check it again.

CHTJonas (changes requested):
Thanks - one tiny pedantic thing but otherwise this seems good!

CHTJonas (approved):

SuperQ (approved):
LGTM

PR #96, opened 2020-09-12
Update readme, sparrc/go-ping to go-ping/ping path

### LINKED ISSUE #90: Call for volunteers: please comment if you are interested in helping maintain go-ping.
There are a variety of open issues and pull requests, most dating back to early 2019 but some as far as 2017. Additionally there have been no commits in over a year.

Are you still interested in maintaining this, @sparrc? If not, would you be amenable to giving commit rights to others?

What is the plan for `go-ping`?

sparrc:
closes #90

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     docker:
       - image: circleci/golang:1.12
-    working_directory: /go/src/github.com/sparrc/go-ping
+    working_directory: /go/src/github.com/go-ping/ping
     steps:
       - checkout
       - run:
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # go-ping
-[![GoDoc](https://godoc.org/github.com/sparrc/go-ping?status.svg)](https://godoc.org/github.com/sparrc/go-ping)
-[![Circle CI](https://circleci.com/gh/sparrc/go-ping.svg?style=svg)](https://circleci.com/gh/sparrc/go-ping)
+[![GoDoc](https://godoc.org/github.com/go-ping/ping?status.svg)](https://godoc.org/github.com/go-ping/ping)
+[![Circle CI](https://circleci.com/gh/go-ping/ping.svg?style=svg)](https://circleci.com/gh/go-ping/ping)
 
 ICMP Ping library for Go, inspired by
 [go-fastping](https://github.com/tatsushid/go-fastping)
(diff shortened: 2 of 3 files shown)


SuperQ (approved):
LGTM

CHTJonas (changes requested):
Guess we should probably change this reference too:

https://github.com/go-ping/ping/blob/22aa99bad7458f4148309129462599c4ceba697c/.circleci/config.yml#L6

SuperQ:
@CHTJonas We can just drop the `working_directory` item from CircleCI configs now.

CHTJonas (approved):

.circleci/config.yml
```diff
@@ -1,23 +0,0 @@
...
-      - run:
-          name: Build and run some basic tests
-          command: |
-            go get ./...
-            go vet ./...
-            go test ./... -cover -race
-            go build -race -o ping_linux ./cmd/ping/ping.go
-            sudo ./ping_linux --privileged -c 2 www.google.com
-            sudo ./ping_linux --privileged -c 3 -i 200ms www.google.com
-            sudo ./ping_linux --privileged -c 10 -i 100ms -t 1s www.google.com
```
sparrc, line 18: re-requesting review because github actions doesn't appear to support this part of the circle ci job, seems like it's because outside internet connectivity is limited within workflow jobs.

SuperQ: Github actions or CircleCI? I'm confused.

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+on: [pull_request]
+
+jobs:
+  unit-tests:
+    name: vet and test
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.12
```
SuperQ, line 12: We should probably use 1.15.

sparrc: should we migrate to go modules first?

.circleci/config.yml
sparrc: Running the ping command was part of the circleci job. Migrating to actions requires us to remove running the ping command.

CHTJonas: I wonder if external ICMP connectivity can be enabled if we get in touch with GitHub support?

SuperQ: I'm wondering why switch from CircleCI to GH actions?

.github/workflows/build.yml
SuperQ: Yea, we can do that as a separate PR.

.circleci/config.yml
sparrc: the only reason was to use github's builtin support, but it may not be worth it if we're missing features

sparrc: okay Ive backed out the github actions changes, perhaps we can revisit at a later time

SuperQ (approved):
LGTM

PR #97, opened 2020-09-12
add github actions

sparrc:


pgollangi:
Hi @sparrc I think [goreleaser](https://goreleaser.com/) can be used to build portable binaries for all platforms. It can be setup in github workflows, perhaps create binaries on creation of tags for release

PR #100, opened 2020-09-14
Add golangci-lint configuration

SuperQ:
* Add a configuration for golangci-lint.
* Fix some linter issues.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -145,7 +145,6 @@ type Pinger struct {
 	addr   string
 
 	ipv4     bool
-	size     int
 	id       int
 	sequence int
 	// network is one of "ip", "ip4", or "ip6".
@@ -309,12 +308,16 @@ func (p *Pinger) Run() {
 		if conn = p.listen(ipv4Proto[p.protocol]); conn == nil {
 			return
 		}
-		conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true)
+		if conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true) != nil {
+			return
+		}
 	} else {
 		if conn = p.listen(ipv6Proto[p.protocol]); conn == nil {
 			return
 		}
-		conn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true)
+		if conn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true) != nil {
+			return
+		}
 	}
 	defer conn.Close()
 	defer p.finish()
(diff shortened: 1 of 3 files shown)


SuperQ:
This is the before report:
```
ping.go:148:2: field `size` is unused (unused)
cmd/ping/ping.go:42:3: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
		fmt.Printf(usage)
		^
cmd/ping/ping.go:62:7: S1005: unnecessary assignment to the blank identifier (gosimple)
		for _ = range c {
		    ^
ping.go:312:42: Error return value of `(*golang.org/x/net/ipv4.PacketConn).SetControlMessage` is not checked (errcheck)
		conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true)
		                                       ^
ping.go:317:42: Error return value of `(*golang.org/x/net/ipv6.PacketConn).SetControlMessage` is not checked (errcheck)
		conn.IPv6PacketConn().SetControlMessage(ipv6.FlagHopLimit, true)
		                                       ^
ping.go:434:24: Error return value of `conn.SetReadDeadline` is not checked (errcheck)
			conn.SetReadDeadline(time.Now().Add(time.Millisecond * 100))
			                    ^
```

ping.go
```diff
@@ -309,12 +308,16 @@ func (p *Pinger) Run() {
 		if conn = p.listen(ipv4Proto[p.protocol]); conn == nil {
 			return
 		}
-		conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true)
+		if conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true) != nil {
+			return
```
CHTJonas, line 312: Would it not be better to capture and return these errors rather than swallowing them?

SuperQ: This function doesn't return anything, so the best it can do is bail. :frowning: 

Changing `recvICMP()` to return an error is something else entirely, so IMO, out of scope for this change.

CHTJonas: What I meant is that modifying `func (p *Pinger) Run()` so that it returns an `error` isn't really a breaking change since the worst that can happen is peoples' linters complain about the ignored return value of `pinger.Run()`. But I take your point that maybe re-engineering how all errors are handled is something for a separate PR.

CHTJonas (approved):

ping.go
SuperQ: https://github.com/go-ping/ping/pull/81 just needs a rebase. :grin:

PR #101, opened 2020-09-14
Rename acronym fields

SuperQ:
Rename `Rtt` and `Ttl` to their respective all-caps names to follow Go
variable naming conventions for acronyms.
* Fix up golint for `Stop()` method.

### DIFF
--- a/cmd/ping/ping.go
+++ b/cmd/ping/ping.go
@@ -70,18 +70,18 @@ func main() {
 
 	pinger.OnRecv = func(pkt *ping.Packet) {
 		fmt.Printf("%d bytes from %s: icmp_seq=%d time=%v ttl=%v\n",
-			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.Rtt, pkt.Ttl)
+			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.RTT, pkt.TTL)
 	}
 	pinger.OnDuplicateRecv = func(pkt *ping.Packet) {
 		fmt.Printf("%d bytes from %s: icmp_seq=%d time=%v ttl=%v (DUP!)\n",
-			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.Rtt, pkt.Ttl)
+			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.RTT, pkt.TTL)
 	}
 	pinger.OnFinish = func(stats *ping.Statistics) {
 		fmt.Printf("\n--- %s ping statistics ---\n", stats.Addr)
 		fmt.Printf("%d packets transmitted, %d packets received, %d duplicates, %v%% packet loss\n",
 			stats.PacketsSent, stats.PacketsRecv, stats.PacketsRecvDuplicates, stats.PacketLoss)
 		fmt.Printf("round-trip min/avg/max/stddev = %v/%v/%v/%v\n",
-			stats.MinRtt, stats.AvgRtt, stats.MaxRtt, stats.StdDevRtt)
+			stats.MinRTT, stats.AvgRTT, stats.MaxRTT, stats.StdDevRTT)
 	}
 
 	pinger.Count = *count
--- a/ping.go
+++ b/ping.go
@@ -29,14 +29,14 @@
 //	}()
 //	pinger.OnRecv = func(pkt *ping.Packet) {
 //		fmt.Printf("%d bytes from %s: icmp_seq=%d time=%v\n",
-//			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.Rtt)
+//			pkt.Nbytes, pkt.IPAddr, pkt.Seq, pkt.RTT)
 //	}
 //	pinger.OnFinish = func(stats *ping.Statistics) {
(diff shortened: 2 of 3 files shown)


SuperQ:
This makes `golint` happy.

sparrc (changes requested):
This is a breaking change, we need to come up with a plan to either make these changes in a v2.0 release tag, or create a "/v2" path for the library that has breaking changes.

SuperQ:
@sparrc We haven't made a tag of any kind yet.

IMO, we should make a `v0.1.0` tag, apply this, then add a `v0.2.0` tag. This is valid within semver.

sparrc:
> @sparrc We haven't made a tag of any kind yet.
> 
> IMO, we should make a `v0.1.0` tag, apply this, then add a `v0.2.0` tag. This is valid within semver, and Go mod versioning.

In semver minor version updates (0.X) are supposed to be backwards compatible, how is this valid?

SuperQ:
That is only after 1.0. For 0.x, there is no requirement.

sparrc:
> That is only after 1.0. For 0.x, there is no requirement.
> 
> The point right after that one is where we're at:
> 
> https://semver.org/#doesnt-this-discourage-rapid-development-and-fast-iteration

I disagree, we're at a stage where this project is being relied on by thousands of individuals and organizations. The next point (https://semver.org/#if-even-the-tiniest-backwards-incompatible-changes-to-the-public-api-require-a-major-version-bump-wont-i-end-up-at-version-4200-very-rapidly) is what I'm advocating for.

We should not make breaking changes to this project lightly. I agree that we should make some breaking changes to bring the library up to a better standard, but I think that we should plan to roll them out together as part of a 2.0 release.

SuperQ:
Also https://semver.org/#spec-item-4

We really are still in initial development here. There are lot of interface issues I would like to fix before we decide to put a 1.0 stamp on it.

sparrc:
> Also https://semver.org/#spec-item-4
> 
> We really are still in initial development here. There are lot of interface issues I would like to fix before we decide to put a 1.0 stamp on it.

The userbase and lifetime (it was started 5 years ago) of this repo contradicts this point.

sparrc:
The other point Im trying to make is that we shouldn't arbitrarily make a decision when we're happy enough to do a 1.0 release. The fact that many people are using and relying on this project is reason enough to make a 1.0 release, and this is something that semver supports (https://semver.org/#how-do-i-know-when-to-release-100)

There are two parts to the answer for "Doesn’t this discourage rapid development and fast iteration?". The answer states that we should be in 0.y.x OR on a development branch working on the next major revision. I would advocate we take the latter approach.

SuperQ:
Go modules and major versions is a _big_ pain to deal with. I would much rather do a couple of small breaking changes in 0.x and get them out of the way.

Go modules already makes it so version pinning works fine in 0.x, there's no need to have a whole bunch of major API versions.

sparrc:
> Go modules and major versions is a _big_ pain to deal with.
> 

how so? is there documentation of that?

SuperQ:
I can't find the discussion thread off the top of my head, but if you look at the [exclude list we maintain](https://github.com/prometheus/prometheus/blob/7de6046d1eb8200fb1b11bd01fc7142bfd3b9a6d/go.mod#L91-L122) in Prometheus to deal with some of the issues.

When it comes to versioning, nobody really cares that much, and from an iteration standpoint, the 0 major version is what we need right now so we can work out what we really want for an API/interface in this library.

sparrc:
> but if you look at the [exclude list we maintain](https://github.com/prometheus/prometheus/blob/7de6046d1eb8200fb1b11bd01fc7142bfd3b9a6d/go.mod#L91-L122) in Prometheus to deal with some of the issues.

You are citing an how onerous it is to maintain an exclude list because these libraries made breaking changes. Does that not suggest to you that making breaking changes to a library should be done more carefully?

SuperQ:
The exclude list is just one of the issues.

What I'm saying is that in order to smoothly deal with a few small breaking changes, staying in the 0 major version is justified. While this code base has been around a long time, there are problems that need to be fixed with breaking changes. Therefore I think the project is not "stable" enough for 1.0. Time does not make code more stable.

CHTJonas (commented):
Reading through the comments so far I think there might be room to compromise here between preserving API compatibility whilst making a lot of breaking changes all at once.

If @sparrc wants to preserve compatibility for anyone currently using the library then I would suggest maybe merging a few of the outstanding PRs that aren't breaking changes and then tagging the resulting commit on the `master` branch as `v1.0.0` which is then promptly ignored and forgotten about. Then in a new commit we bump the module version by changing the `go.mod` header to `module github.com/go-ping/ping/v2`. I do agree with @SuperQ that the codebase is in need of a fair bit of TLC so a new major version gives us somewhere to ~break~ improve things and rapidly iterate. When we're happy with where things are we can then tag the new master state as `v2.0.0` and maybe make a note in the README that people should update to v2.

The [recommended way](https://blog.golang.org/v2-go-modules) for libraries to work on a new breaking version is to create a `v2/` or `v3/` directory within their project structure which I think would be _massively_ OTT for us here; git tags are probably fine and much easier than keeping around two copies of everything. I think the only downside to this is that people who are not using go modules won't be able to upgrade - should we care?

PR #103, opened 2020-09-18
Adds a `logger` interface and `SetLogger` function

### LINKED ISSUE #16: Library should return errors and not printf messages

maddie:
Closes #16

### DIFF
--- a/cmd/ping/ping.go
+++ b/cmd/ping/ping.go
@@ -33,6 +33,15 @@ Examples:
     sudo ping --privileged www.google.com
 `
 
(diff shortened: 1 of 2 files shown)


SuperQ:
Sorry, this needs a rebase.

PR #107, opened 2020-10-01
Fix `recvICMP` return arguments

s-newman:
The `Pinger.recvICMP()` method should return `nil` instead of returning nothing. The current return statement causes issues when attempting to `go get` this package.

Running:
```
go get github.com/go-ping/ping
```

Causes the following error:
```
# github.com/go-ping/ping
../../../go/pkg/mod/github.com/go-ping/ping@v0.0.0-20201001152913-70ede2ab323b/ping.go:476:5: not enough arguments to return
	have ()
	want (error)
```

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -473,7 +473,7 @@ func (p *Pinger) recvICMP(
 
 			select {
 			case <-p.done:
-				return
+				return nil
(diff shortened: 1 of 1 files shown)


s-newman:
Whoops, this is a duplicate of #106. My mistake.

SuperQ:
No worries, thanks for hopping on the problem. :smile:

Issue #105, opened 2020-09-23
not implemented on windows/amd64


qin-nz:
version: `github.com/go-ping/ping v0.0.0-20200918120429-e8ae07c3cec8`

Run following code on Windows 10, it says `not implemented on windows/amd64`.

```
package main

import (
	"fmt"

	"github.com/go-ping/ping"
)

func main() {
	pinger, err := ping.NewPinger("180.76.76.76")
	pinger.SetPrivileged(true)
	pinger.Count = 3
	err = pinger.Run()
	if err != nil {
		fmt.Println(err)
	}
}

```

houaq:
Same problem +1!

houaq:
The error is caused by `SetControlMessage` call, It seems like to check the packet TTL. I solved this by comment it out, not sure about other side effects.

c0decave:
> The error is caused by `SetControlMessage` call, It seems like to check the packet TTL. I solved this by comment it out, not sure about other side effects.
Interesting, at my side the linux version crashes, the windows version still throws the same error message :(

CHTJonas:
I appears this is a limitation in the Go `x/net/ipv{4|6}` packages:

- https://pkg.go.dev/golang.org/x/net/ipv4#pkg-note-BUG
- https://pkg.go.dev/golang.org/x/net/ipv6#pkg-note-BUG

https://github.com/go-ping/ping/blob/671c40f29adc8190ded8c57bbebf9b6ea534c7ea/ping.go#L452-L454

Since we're already checking whether the `ControlMessage` is present or not I think it's safe to simply pass over the `conn.IPv4PacketConn().SetControlMessage(ipv4.FlagTTL, true)` code lines on Windows, ditto IPv6.

PR #108, opened 2020-10-06
Ignore control message errors on Windows

CHTJonas:
On Windows, the Go `x/net/ipv4` and `x/net/ipv6` packages do not implement the necessary control message functionality for us to access packet TTLs (see [here](https://pkg.go.dev/golang.org/x/net/ipv4#pkg-note-BUG) and [here](https://pkg.go.dev/golang.org/x/net/ipv6#pkg-note-BUG)). This PR ignores any control message errors that are returned by the `SetControlMessage` function if the Go runtime is compiled targeting Windows.

This fixes issue #105.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -51,6 +51,7 @@ import (
 	"math"
 	"math/rand"
 	"net"
+	"runtime"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
LGTM, Thanks!

sparrc (approved):
thanks for digging into that!

PR #109, opened 2020-10-06
Add EditorConfig file

CHTJonas:
[EditorConfig](https://editorconfig.org) is a dotfile which stores configuration that is supported by a wide variety of text editors and IDEs both natively and with plugins. The advantage of this is that it allows us to enforce some project standards such as the style of indentation and line ending used. I find this can help reduce git conflicts in some circumstances.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# https://editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+
+[Makefile]
+indent_style = tab
+indent_size = 4
+
+[*.go]
+indent_style = tab
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,8 +11,6 @@ indent_style = space
 
 [Makefile]
 indent_style = tab
-indent_size = 4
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+
+[Makefile]
+indent_style = tab
+indent_size = 4
```
SuperQ, line 14: I would prefer to keep tab size opinions up to the editor.

SuperQ (commented):
I'm not sure I'm in favor of this or not. I usually prefer to keep editor settings opinions out of the codebase. However ,I do like the standardization of things like final newlines and trim trailing whitespace.

Normally I would suggest these should be enforced by CI, not editor settings. But I suppose using this config is still voluntary.

.editorconfig
CHTJonas: No problem - will remove from config 👍

SuperQ:
Yes, I think this is a good first step in getting it to where it needs to be.

SuperQ (approved):
LGTM

sparrc (approved):
TIL about editorconfig, lgtm

PR #114, opened 2020-10-18
Add an OnSend handler.

rrcollier:
Similar to the OnReceive handler but called when a packet is sent. Reuses the Packet type, but does not set Ttl or Rtt.

This was originally accidentally included in PR 104. Now seperated out.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -124,6 +124,9 @@ type Pinger struct {
 	// rtts is all of the Rtts
 	rtts []time.Duration
 
+	// OnSend is called when Pinger sends a packet
+	OnSend func(*Packet)
+
 	// OnRecv is called when Pinger receives and processes a packet
 	OnRecv func(*Packet)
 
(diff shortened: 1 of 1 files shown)


ping.go
```diff
@@ -587,6 +590,17 @@ func (p *Pinger) sendICMP(conn *icmp.PacketConn) error {
...
+
+		handler := p.OnSend
+		if handler != nil {
+			outPkt := &Packet{
+				Nbytes: len(msgBytes),
+				IPAddr: p.ipaddr,
+				Addr:   p.addr,
+				Seq:    p.sequence - 1,
+			}
+			handler(outPkt)
```
CHTJonas, line 602: I did wonder whether this should be asynchronous `go handler(outPkt)` but it matches the other `OnRecv` event handlers we have in the code. We can look at this in a future PR perhaps.

CHTJonas (approved):
~LGTM 👍~ Edit: see review below.

ping.go
```diff
@@ -587,6 +590,17 @@ func (p *Pinger) sendICMP(conn *icmp.PacketConn) error {
...
+		}
```
CHTJonas, line 603: If you swap around the order of execution it avoid doing an unnecessary `p.sequence - 1`:

```suggestion
		handler := p.OnSend
		if handler != nil {
			outPkt := &Packet{
				Nbytes: len(msgBytes),
				IPAddr: p.ipaddr,
				Addr:   p.addr,
				Seq:    p.sequence,
			}
			handler(outPkt)
		}

		p.PacketsSent++
		p.sequence++
```

CHTJonas (changes requested):

ping.go
rrcollier: Yes, I modeled the behaviour after the OnRecv handler.

SuperQ (approved):
LGTM

CHTJonas (approved):

PR #115, opened 2020-10-18
Add option not to store all received rtts.

rrcollier:
This prevents memory bloat from a process that wants to do very long running ping tests.

This was previously included by accident in PR 104.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -121,6 +121,10 @@ type Pinger struct {
 	// Number of packets received
 	PacketsRecv int
 
+	// If true, keep a record of rtts of all received packets.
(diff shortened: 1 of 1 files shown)


SuperQ:
This will default to off, which will make the `Statistics()` function not work by default because of the way it uses the `rtts` slice.

This should probably be enabled in the `New()` method.

rrcollier:
Fixed. Thanks for spotting that.

SuperQ (approved):
LGTM.

CHTJonas (approved):
🚀

PR #83, opened 2020-03-22
Close the go channel gracefully

voidint:
golang repeatedly closing the channel will result in panic. 

```go
pinger, err := ping.NewPinger("www.google.com")
if err != nil {
        panic(err)
}
defer pinger.Stop() // panic
```

PR #116, opened 2020-10-18
Allow pinger.Stop() to be safely called repeatedly

### LINKED ISSUE #94: Pinger.Stop() is unsafe when used with a timeout or count
Calling the `Stop()` method on a pinger with a timeout or count configured can cause a panic, if the `Stop()` is called after the Pinger has exceeded either the timeout or ping count:

```
panic: close of closed channel

goroutine 10 [running]:
github.com/sparrc/go-ping.(*Pinger).Stop(...)
	/go/pkg/mod/github.com/sparrc/go-ping@v0.0.0-20190613174326-4e5b6552494c/ping.go:345
```

There's no way to safely prevent this panic externally without some kind of race condition.

CHTJonas:
This PR is an extensions to #83 and fixes #94.

It synchronises `pinger.Stop()` and makes it safe to call repeatedly. It also changes the return signature of the method so that it returns a 'benign' (safe to ignore) error if called more than once.

I'm not sure what your opinions are on returning errors on subsequent method calls - if you think it's a bit over-the-top then I think we can use a `sync.Once` as #83 does instead which is probably a simpler solution.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -69,8 +69,9 @@ const (
 )
 
 var (
-	ipv4Proto = map[string]string{"icmp": "ip4:icmp", "udp": "udp4"}
-	ipv6Proto = map[string]string{"icmp": "ip6:ipv6-icmp", "udp": "udp6"}
+	ipv4Proto   = map[string]string{"icmp": "ip4:icmp", "udp": "udp4"}
+	ipv6Proto   = map[string]string{"icmp": "ip6:ipv6-icmp", "udp": "udp6"}
+	ErrTornDown = errors.New("Stop() called more than once")
 )
 
 // New returns a new Pinger struct pointer.
@@ -84,7 +85,7 @@ func New(addr string) *Pinger {
 		Tracker:  r.Int63n(math.MaxInt64),
 
 		addr:     addr,
-		done:     make(chan bool),
+		done:     make(chan interface{}),
 		id:       r.Intn(math.MaxInt16),
 		ipaddr:   nil,
 		ipv4:     false,
@@ -140,8 +141,9 @@ type Pinger struct {
 	// Source is the source IP address
 	Source string
 
-	// stop chan bool
-	done chan bool
+	// Cleardown channel and mutex
+	done chan interface{}
+	lock sync.Mutex
 
 	ipaddr *net.IPAddr
 	addr   string
@@ -350,7 +352,7 @@ func (p *Pinger) Run() error {
 			wg.Wait()
 			return nil
 		case <-timeout.C:
-			close(p.done)
+			_ = p.Stop()
 			wg.Wait()
 			return nil
 		case <-interval.C:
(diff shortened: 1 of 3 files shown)


ping.go
```diff
@@ -370,15 +372,22 @@ func (p *Pinger) Run() error {
...
 		}
 	}
 }
 
-func (p *Pinger) Stop() {
+func (p *Pinger) Stop() error {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if p.torndown {
+		return errors.New("Stop() called more than once")
```
SuperQ, line 386: Let's define this as a package variable so that it can be tested with `errors.Is(err, ping.ErrTornDown)`.

```
var ErrTornDown = errors.New("Stop() called more than once")
```

ping.go
```diff
@@ -140,8 +140,10 @@ type Pinger struct {
 	// Source is the source IP address
 	Source string
 
-	// stop chan bool
-	done chan bool
+	// Cleardown channel, state var and mutex
+	done     chan interface{}
+	torndown bool
```
SuperQ, line 145:
```suggestion
	tornDown bool
```

SuperQ (commented):
While this works, maybe there's a more Go ideomatic way to tell if a channel is closed so we avoid double closing it.

CHTJonas:
Thanks for the feedback - should be good for you to have another look now!

The tests I added didn't work, I think because CircleCI runner's doesn't let you bind to the necessary sockets. Not sure if it's worth (in a separate PR) making another test suite which people can run locally which actually pings some real endpoints? Or maybe we can email Circle's support and ask if the socket functionality can be enabled somehow?

SuperQ (approved):
LGTM, Nice.

SuperQ:
Yea, I'm not sure about how to improve tests in CircleCI. These are the kinds of things that are too easy to abuse without more isolation in the containers that CircleCI runs.

SuperQ:
Would you mind squashing/cleaning up the commits for this PR?

ping.go
```diff
@@ -370,15 +372,29 @@ func (p *Pinger) Run() error {
...
+	defer p.lock.Unlock()
+
+	ok := true
+	select {
+	case _, ok = <-p.done:
+	default:
+	}
+
+	if !ok {
+		return ErrTornDown
```
sparrc, line 393: What would be the use-case of this error? seems like most people would want to ignore this situation, which is a bit easier and cleaner if we just return `nil`

ping.go
```diff
@@ -370,15 +372,29 @@ func (p *Pinger) Run() error {
...
-			close(p.done)
+			_ = p.Stop()
 			wg.Wait()
 			return nil
 		}
 	}
 }
 
-func (p *Pinger) Stop() {
+func (p *Pinger) Stop() error {
```
sparrc, line 382: if we do end up returning ErrTownDown when Stop() has already been called, we should document that behavior in a comment on this function

ping.go
CHTJonas: The `error` returned is just an indicator really that I thought _might_ be useful to someone somewhere. It could equally be a true/false boolean flag. Or nothing, as you suggest - I'm not dead set on it by any means. I could perhaps see it as being useful in situations where people are using a pinger in goroutine and might like to use it to help detect race conditions.

PR #117, opened 2020-10-26
Add Reset method to be able to re-use the Pinger instance

zekroTJA:
I have a specific use case where I want to periodically determine the RTT to a server. 

I tried to re-use the `Pinger` instance and used `Run` on every ticker elapse on the same instance. The result was that every following `Statistics` instance contained the same RTT times. Below, you can find the snippet for this.
https://gist.github.com/zekroTJA/ae3dbbe2dbbe56b3c1be9775f77718f5

To be able to re-use the `Pinger` instance, I have added a `Reset` method which reinitializes the `done` channel and the `rtts` array and sets the values of `PacketsSent` and `PacketsRecv` back to zero.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -518,6 +518,16 @@ func (p *Pinger) Stop() {
 	}
 }
 
+// Reset sets the RTTs and the sent and received counts
+// back to zero so that Run can be re-used on the same
+// instance.
+func (p *Pinger) Reset() {
+	p.done = make(chan bool)
+	p.rtts = make([]time.Duration, 0)
+	p.PacketsSent = 0
+	p.PacketsRecv = 0
+}
+
 func (p *Pinger) finish() {
 	handler := p.OnFinish
 	if handler != nil {


SuperQ:
If you want to track the RTT to a server, maybe it's easier to make your own hook into OnRecv. This is what I do in the [smokeping_prober](https://github.com/SuperQ/smokeping_prober/blob/9ba85274dcc21bf8132cbe3b3dccfcb4aab57d9f/collector.go#L66-L71).

CHTJonas:
@zekroTJA could you rebase this on top of the latest `master` please?

While I'm not sure the code in your gist is necessarily the best way to go about achieving want I think you're trying to, I can see the use of a `Reset()` method in other situations; it could be very useful for people who want to reuse a pinger instance many times in order to avoid GC churn.

CHTJonas (changes requested):
@zekroTJA could you rebase this on top of the latest `master` please?

While I'm not sure the code in your gist is necessarily the best way to go about achieving want I think you're trying to, I can see the use of a `Reset()` method in other situations; it could be very useful for people who want to reuse a pinger instance many times in order to avoid GC churn.

CHTJonas:
@zekroTJA, please rebase!

SuperQ:

PR #119, opened 2020-11-05
Add option to ping forever

tydra-wang:
If Pinger.Timeout is set to nonpositive duration, ping forever.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /ping
 /dist
+.idea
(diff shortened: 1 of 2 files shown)


SuperQ:
This is mostly unnecessary, setting `p.Timeout = time.Duration(math.MaxInt64)` is 290 years. This is effectively forever.

CHTJonas:
@tydra-wang I'm afraid we've decided to resolve this in #120 instead as that introduces less complexity into the code base but thank you for the contribution nevertheless!

PR #120, opened 2020-11-06
Make default Timeout "forever"

SuperQ:
Make the default Timeout setting the maximum Go Duration, which is
290 years. Effectively forever.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -81,7 +81,7 @@ func New(addr string) *Pinger {
 		Interval:   time.Second,
 		RecordRtts: true,
 		Size:       timeSliceLength,
-		Timeout:    time.Second * 100000,
+		Timeout:    time.Duration(math.MaxInt64),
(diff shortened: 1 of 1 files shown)


SuperQ:
Alternative approach to https://github.com/go-ping/ping/pull/119.

CHTJonas (approved):
LGTM 🚀

We probably want to update the timeout in the demo app as well:

https://github.com/go-ping/ping/blob/06a3dc498947cb83911661822b1d79eae5e7dd1f/cmd/ping/ping.go#L37

sparrc (approved):

SuperQ:
@CHTJonas I'm not sure there's much need to update the demo app. It's a nice demonstration of how to use the feature.

CHTJonas:
Yeah, that's actually a good point. Happy with where this is 💯

Issue #124, opened 2020-11-27
Should the `Size` in function `New` be `timeSliceLength+trackerLength` ?


chenlujjj:
In function `sendICMP`:
```
	t := append(timeToBytes(time.Now()), intToBytes(p.Tracker)...)
	if remainSize := p.Size - timeSliceLength - trackerLength; remainSize > 0 {
		t = append(t, bytes.Repeat([]byte{1}, remainSize)...)
	}

	body := &icmp.Echo{
		ID:   p.id,
		Seq:  p.sequence,
		Data: t,
	}
```
so `len(t)=16`, not p.Size which is 8。

When run command: `go run cmd/ping/ping.go -c 5 -i 200ms  www.baidu.com `, output is:
```
PING www.baidu.com (180.101.49.12):
24 bytes from 180.101.49.12: icmp_seq=0 time=10.268ms ttl=51
24 bytes from 180.101.49.12: icmp_seq=1 time=8.643ms ttl=51
24 bytes from 180.101.49.12: icmp_seq=2 time=8.947ms ttl=51
24 bytes from 180.101.49.12: icmp_seq=3 time=9.439ms ttl=51
^C
--- www.baidu.com ping statistics ---
4 packets transmitted, 4 packets received, 0% packet loss
round-trip min/avg/max/stddev = 8.643ms/9.32425ms/10.268ms/614.461µs
```
Received packet size is 24 bytes, so the sent packet size is 16 bytes.

If I misunderstand something here, please point out. Thanks.

PaulChenAU:
My understanding is yes since ```p.Size``` equals to 8 which is timeSliceLength as shown in the constructor.
```go
if remainSize := p.Size - timeSliceLength - trackerLength; remainSize > 0 {
    t = append(t, bytes.Repeat([]byte{1}, remainSize)...)
}
```
and the code inside this if statement will never get executed in my opinion.   

Actually, you can refer to the ```go-fastping``` repository as it sends ```icmp.Echo``` message and transforms current timestamp to bytes array with length 8 as the data body. So ```go-fastping``` set p.Size = 8 which is timeSliceLength .  
  
However, this ping repo uses the current timestamp as well as the unique tracker int as the data body. Therefore, I think p.Size inside the ```New()``` constructor should add trackerLength.

CHTJonas:
[...]

PR #127, opened 2020-12-25
add trackerLength to p.Size in New()

scientiacoder:
Hi, there  

Refer to the ```go-fastping``` repository as it sends ```icmp.Echo``` message and transforms current timestamp to bytes array with length 8 as the data body. So ```go-fastping``` set p.Size = 8 which is timeSliceLength in constructor.

However, this ```ping``` repo uses the current timestamp as well as the unique tracker int as the data body. Therefore, I think ```p.Size``` inside the ```New()``` constructor should add trackerLength.  

issue #124

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -89,7 +89,7 @@ func New(addr string) *Pinger {
 		Count:      -1,
 		Interval:   time.Second,
 		RecordRtts: true,
-		Size:       timeSliceLength,
+		Size:       timeSliceLength + trackerLength,
 		Timeout:    time.Second * 100000,
 		Tracker:    r.Int63n(math.MaxInt64),
 


CHTJonas (commented):
Thanks for this! Would you be able to add a test case for this change? Worth noting that our current suite has this line which seems odd and might want deleting:

https://github.com/go-ping/ping/blob/30a8f08ad2a9d0b88ca9c1978114d253f63748c3/ping_test.go#L469

CHTJonas:
I'm now wondering whether the default `Size` in the `NewPinger()` method should be 56 to match the default that `/bin/ping` uses. Maybe that would be more standard and What People Expect™?

scientiacoder:
Did some research on the **size** of the ```ping``` command:  
Both **Unix** and **macOS** uses 56 as the default data length, but they have different ```fill()``` methods
1. Unix seems like using a for loop to fill and pad ICMP echo data payload
```c
// Source: https://nsrc.org/wrc/materials/src/ping.c

static u_char outpack[MAXPACKET];
register struct icmp *icp = (struct icmp *) outpack;
int i, cc;
register struct timeval *tp = (struct timeval *) &outpack[8];
register u_char *datap = &outpack[8+sizeof(struct timeval)];
icp->icmp_type = ICMP_ECHO;
icp->icmp_code = 0;
icp->icmp_cksum = 0;
icp->icmp_seq = ntransmitted++;
icp->icmp_id = ident;		/* ID */

cc = datalen+8;			/* skips ICMP portion */

if (timing)
	gettimeofday( tp, &tz );

for( i=8; i<datalen; i++)	/* skip 8 for time */ /* this for loop padding the data */
	*datap++ = i;

/* Compute ICMP checksum here */
icp->icmp_cksum = in_cksum( icp, cc );

/* cc = sendto(s, msg, len, flags, to, tolen) */
i = sendto( s, outpack, cc, 0, &whereto, sizeof(struct sockaddr) );
```
2. macOS looks like padding **hexadecimal unsigned integers**(maybe)
```c
// Source: https://opensource.apple.com/source/network_cmds/network_cmds-328/ping.tproj/ping.c

fill(){
...
	for (cp = patp; *cp; cp++) {
		if (!isxdigit(*cp))
			errx(EX_USAGE,
			    "patterns must be specified as hex digits");

	}
	ii = sscanf(patp,
	    "%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x%2x",
	    &pat[0], &pat[1], &pat[2], &pat[3], &pat[4], &pat[5], &pat[6],
	    &pat[7], &pat[8], &pat[9], &pat[10], &pat[11], &pat[12],
	    &pat[13], &pat[14], &pat[15]);

	if (ii > 0)
		for (kk = 0; kk <= maxpayload - (TIMEVAL_LEN + ii); kk += ii)
			for (jj = 0; jj < ii; ++jj)
				bp[jj + kk] = pat[jj];
```
or maybe **Linux** uses hex too, packet capture from Wireshark:
image
  
So if it is really more standard and What People Expect™, currently ```go-ping``` uses []byte{1} to **fill** ICMP echo data payload, maybe this also needs to be changed? 
  
I think another option may be like add a new **flag** -s indicates packet size, like the ```ping command [-s packetsize]```
```
ping -s 24 github.com
```
But the size range should be considered and it seems like in *nix the ping statistics will **not be** calculated when the size range is too small like 0~15(maybe for size 0, can't put a timestamp in it[Doge]):
image

CHTJonas:
FWIW when you set `pinger.Size` you're only setting the size of the message _body_. ICMP messages have an 8-byte header so that's why you see the "8 bytes from XXX" log lines and that's to be expected. Apologies if I'm preaching to the choir! I think a `-s` flag for the ping command is a good idea for testing etc. though we don't aim to _fully_ replicate all the functionality in the `/bin/ping` binaries that tend to come with the *NIXes.

I agree that the current method of using `[]byte{1}` could be a bit better: a message payload that isn't padded with just binary ones would be better for testing echoing and packet (de)serialisation of each host in the packet's path. @SuperQ mentioned that maybe we could encode a 16-byte UUID in place of the current tracker and then truncate that if the user wants a smaller messages size or add it somehow if they want a larger one.

scientiacoder:
Haha, thanks for your reply. Even do want to replicate ```ping``` command, it looks like operating systems do not have a standard ping implementation.

CHTJonas:
Looks like merging #138 has broken `master` if a `pinger.Size` hasn't been manually set (i.e. because the default size is wrong the deserialisation is broken). Let's get this merged ASAP so master works again and maybe we can pickup the related stuff about changing the padding algorithm and encoding UUIDs in different PRs (perhaps).

CHTJonas (approved):

sparrc (approved):

PR #130, opened 2021-01-28
Process duplicate packets and make seeds goroutine-safe

### LINKED ISSUE #14: PacketLoss maybe negative
4 packets transmitted, 5 packets received, -25% packet loss

### LINKED ISSUE #129: Running multiple pingers in concurrent causes true ping on wrong ip
Hello,

I'm using this package for a project and I have a problem when I execute multiple pings.
I have 11 ping in concurrent and I only have three reachable IPs out of the eleven.

When i execute my go routine i have some ip out which have a true state.
The pings are random over time.
I tested with a bash ping exec directly and I don't have this problem.

Why ?

OS Running : Linux Ubuntu 20.04 and 18.04 / Windows 10
Go : 1.15.6 and 1.15.17

Code to catch true ping :
`
if state.state == true && target != "8.8.8.8" && target != "10.0.202.50" && target != "10.0.202.51" {
 log.Error(target, " : ", state.state)
}
`
Log : 
> ERRO[0028] 10.0.202.57 : true
> ERRO[0046] 10.0.202.52 : true
> ERRO[0088] 10.0.202.54 : true
> ERRO[0096] 10.0.202.58 : true
> ERRO[0100] 10.0.202.52 : true

@CHTJonas

ssoroka:
We were seeing some issues with packets being delivered multiple times, suspecting duplicate sequence numbers resulting in duplicate counting of received packets and negative loss percentage.

Cleaned up some code around unique IDs and potentially overlapping IDs resulting from bad random number seeds.

- ignores second delivery of same sequence number.
- make seeds goroutine-safe so we never have a duplicate seed
- expand id to int32 rather than int16 so that we never get collisions, and there was space to do this anyway.

Fixes #14

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -134,6 +134,9 @@ type Pinger struct {
 	// Number of packets received
 	PacketsRecv int
 
+	// Number of duplicate packets received
+	PacketsRecvDuplicates int
+
 	// If true, keep a record of rtts of all received packets.
 	// Set to false to avoid memory bloat for long running pings.
 	RecordRtts bool
@@ -149,6 +152,9 @@ type Pinger struct {
 
 	// OnFinish is called when Pinger exits
 	OnFinish func(*Statistics)
+
+	// OnDuplicateRecv is called when a packet is received that has already been received.
+	OnDuplicateRecv func(*Packet)
 
 	// Size of packet being sent
 	Size int
@@ -211,6 +217,9 @@ type Statistics struct {
 
 	// PacketsSent is the number of packets sent.
 	PacketsSent int
+
+	// PacketsRecvDuplicates is the number of duplicate responses there were to a sent packet.
+	PacketsRecvDuplicates int
 
 	// PacketLoss is the percentage of packets lost.
 	PacketLoss float64
@@ -430,14 +439,15 @@ func (p *Pinger) Statistics() *Statistics {
 		total += rtt
 	}
 	s := Statistics{
-		PacketsSent: p.PacketsSent,
-		PacketsRecv: p.PacketsRecv,
-		PacketLoss:  loss,
-		Rtts:        p.rtts,
-		Addr:        p.addr,
-		IPAddr:      p.ipaddr,
-		MaxRtt:      max,
-		MinRtt:      min,
+		PacketsSent:           p.PacketsSent,
+		PacketsRecv:           p.PacketsRecv,
+		PacketsRecvDuplicates: p.PacketsRecvDuplicates,
+		PacketLoss:            loss,
+		Rtts:                  p.rtts,
+		Addr:                  p.addr,
+		IPAddr:                p.ipaddr,
+		MaxRtt:                max,
+		MinRtt:                min,
 	}
 	if len(p.rtts) > 0 {
 		s.AvgRtt = total / time.Duration(len(p.rtts))
@@ -555,6 +565,10 @@ func (p *Pinger) processPacket(recv *packet) error {
 		outPkt.Seq = pkt.Seq
 		// If we've already received this sequence, ignore it.
 		if _, inflight := p.awaitingSequences[pkt.Seq]; !inflight {
+			p.PacketsRecvDuplicates++
+			if p.OnDuplicateRecv != nil {
+				p.OnDuplicateRecv(outPkt)
+			}
 			return nil
 		}
 		// remove it from the list of sequences we're waiting for so we don't get duplicates.
--- a/ping.go
+++ b/ping.go
@@ -96,7 +96,7 @@ func New(addr string) *Pinger {
 
 		addr:              addr,
 		done:              make(chan bool),
-		id:                r.Intn(math.MaxInt32),
+		id:                r.Intn(math.MaxInt16),
 		ipaddr:            nil,
 		ipv4:              false,
 		network:           "ip",
--- a/ping.go
+++ b/ping.go
@@ -553,7 +553,7 @@ func (p *Pinger) processPacket(recv *packet) error {
 
 		outPkt.Rtt = receivedAt.Sub(timestamp)
 		outPkt.Seq = pkt.Seq
-		// if we've already received this sequence, ignore it.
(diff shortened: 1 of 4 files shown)


CHTJonas:
Thanks for this PR @ssoroka! Will give it a look over and play around. As an aside this *may* fix #129 but we're waiting an exact reproducible test case for that.

ssoroka:
I've addressed the memory use issue by making the map only use memory for tracking in-flight sequences that weren't delivered, so that under normal operation it uses zero extra memory.
It'd be more exact to track how long we've been waiting for a sequence and stop waiting for delivery, freeing that memory, but it makes the code more complex and I think this is a reasonable trade-off for simplicity. Open to thoughts.

CHTJonas (changes requested):
Hello again 👋  This all looks great except that it seems to be broken, at least for me on macOS and Linux! The change of `id` from `r.Intn(math.MaxInt16)` to `r.Intn(math.MaxInt32)` seems to cause packets to be lost. I haven't investigated this much further (e.g. with a packet sniffer) than by testing with the code below:

```bash
git clone https://github.com/go-ping/ping.git
cd ping
sudo go run cmd/ping/ping.go --privileged -c 4 www.google.com
# 24 byte replies :-)
curl -sS https://patch-diff.githubusercontent.com/raw/go-ping/ping/pull/130.diff | git apply
sudo go run cmd/ping/ping.go --privileged -c 4 www.google.com
# no replies :-(
sed -e 's/math.MaxInt32/math.MaxInt16/' -I ping.go
sudo go run cmd/ping/ping.go --privileged -c 4 www.google.com
# working again! :-D
```

Could you revert back to `math.MaxInt16` and maybe open another separate PR that just addresses changing that? Where possible we always prefer a single PR to fix/enhance a single 'thing'.

ssoroka:
Sure, I'm happy to pull that change out. I do find it very strange that this causes packet loss though. I was worried that 2^15 (one bit sign) is only 32768, and so for someone running, say, 100 pings in parallel (as I have a user doing), you'd actually have a ~1/300 chance of ID collision, pretty much guaranteeing an ID collision every 5 minutes. I'm guessing the tracking ID avoids this, but I wasn't 100% sure that was the case. It seems like we should be using the full 31 bits here.

ssoroka:
I was able to replicate the issue with MaxInt32, and I've reverted it to MaxInt16 as it was before.

ping.go
```diff
@@ -549,6 +553,12 @@ func (p *Pinger) processPacket(recv *packet) error {
 
 		outPkt.Rtt = receivedAt.Sub(timestamp)
 		outPkt.Seq = pkt.Seq
+		// if we've already received this sequence, ignore it.
```
SuperQ, line 556: Nit, please make full sentences.

```suggestion
		// If we've already received this sequence, ignore it.
```

CHTJonas (dismissed):

CHTJonas:
Happy with where this is now! I'm testing the code from this PR myself in a reasonably concurrent environment and all seems fine -- no observable increases in CPU or memory usage. 🚀

SuperQ:
We should probably get some benchmark tests going for some of these changes.

SuperQ:
I don't think we actually want to ignore duplicate packets, this breaks duplicate packet reporting.

CHTJonas:
I was going to propose adding `stats.PacketsDup` to `ping.Statistics` and/or maybe a `pinger.OnDup` callback for that purpose but in a different PR. To be honest I think it's better to get duplicate rejection in sooner rather than later but we could do that here.

ssoroka:
hmm. if we actually want to report on duplicates, we should probably do a better job of it, as the current results are confusing at the least. the loss percentages can be negative and you can get weird results when pinging multicast endpoints where you might get multiple responders. i don’t think anyone is expecting to receive more replies than they sent, so this is really shocking behavior; also consider sending 3 pings where ping 1 gets 1 reply, ping 2 gets 0 replies, and ping 3 gets 2 replies: this is not zero packet loss. 

as far as i know, i’ve also never seen other ping libraries acting like you suggest.

SuperQ:
Yes, having a separate counter/handler for duplicates is probably better.

There's already an open PR to attempt to record DUPs. https://github.com/go-ping/ping/pull/84

ssoroka:
actually i'm happy to edit the pr to include better duplicate handing: the callback you suggest and return it as part of the ping statistics

CHTJonas:
@ssoroka go for it 👍

ssoroka:
This now closes https://github.com/go-ping/ping/pull/84 as well.

ssoroka:
let me know if you prefer the shortened PacketsDups vs PacketsRecvDuplicates

ping.go
```diff
@@ -134,6 +134,9 @@ type Pinger struct {
 	// Number of packets received
 	PacketsRecv int
 
+	// Number of duplicate packets received
+	PacketsRecvDuplicates int
```
SuperQ, line 138: Should we make the counters explicit uint64 going forward? Or maybe make it a breaking change later?

ssoroka: I thought about that, but PacketsRecv and others right above it are already ints, so it seemed weird to have ints and then one uint64

CHTJonas: I think I too would go for consistency here. Not opposed to changing them all at once elsewhere but it seems a bit out of scope for this PR.

CHTJonas (approved):
Thanks! This is looking good IMO :shipit: I was a bit cheeky and took the liberty of tending to the nit that is updating the README and the included 'demo' app under the `cmd` directory myself rather than requesting changes again -- hope you don't mind.

CHTJonas:
FYI we'll probably squash this when we merge just to keep the history tidy.

ssoroka:
I think I'm done here and this is good to squash-merge from my end. Thanks for the assistance!

ping.go
SuperQ: Consistency for now is fine.

SuperQ (approved):

PR #134, opened 2021-02-01
Make Tracker uint64 and check for both UDP and ICMP

### LINKED ISSUE #133: Use "udp" in multi ping situation, respose is error for not checking the id.
When I use "udp" protocol in multi ping situation. I would be error for the code below,because the protocol is "udp", It wouldn't check the id which makes other ping response also go through the left code .
processPacket:
case *icmp.Echo:
		// If we are priviledged, we can match icmp.ID
		if p.protocol == "icmp" {
			// Check if reply from same ID
			if pkt.ID != p.id {
				return nil
			}
		}

CHTJonas:
It looks like we can check a received packet's `ID` in both UDP and ICMP so we might as well. This PR also reduces some needless complexity by making `Tracker` an unsigned 64-bit int.

Fixes #133.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -92,7 +92,7 @@ func New(addr string) *Pinger {
 		RecordRtts: true,
 		Size:       timeSliceLength,
 		Timeout:    time.Second * 100000,
-		Tracker:    r.Int63n(math.MaxInt64),
+		Tracker:    r.Uint64(),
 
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
LGTM

sparrc (approved):

PR #145, opened 2021-02-09
Add size flag to demo pinger

CHTJonas:
This PR adds a command line flag to the demo pinger which lets the user set the the size of the ICMP payload. The default is
set at 16 which is the current default for an unconfigured pinger. Mainly useful for testing purposes.

### DIFF
--- a/cmd/ping/ping.go
+++ b/cmd/ping/ping.go
@@ -31,12 +31,16 @@ Examples:
 
     # Send a privileged raw ICMP ping
     sudo ping --privileged www.google.com
+
+    # Send ICMP messages with a 100-byte payload
+    ping -s 100 1.1.1.1
 `
 
 func main() {
 	timeout := flag.Duration("t", time.Second*100000, "")
 	interval := flag.Duration("i", time.Second, "")
 	count := flag.Int("c", -1, "")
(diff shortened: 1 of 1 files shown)


cmd/ping/ping.go
```diff
@@ -31,12 +31,16 @@ Examples:
...
+
+    # Send ICMP messages with a 100-byte payload
+    ping -s 100 1.1.1.1
 `
 
 func main() {
 	timeout := flag.Duration("t", time.Second*100000, "")
 	interval := flag.Duration("i", time.Second, "")
 	count := flag.Int("c", -1, "")
+	size := flag.Int("s", 16, "")
```
SuperQ, line 43: Default to traditional ping 56?

CHTJonas: I was thinking of keeping it at the current default for now and then opening a separate PR to propose we change it to the traditional default both here and in the following line:

https://github.com/go-ping/ping/blob/2c1f7d31c4d0a9d40be1106d8e0274ee718738c9/ping.go#L93

SuperQ (approved):

sparrc (approved):

PR #147, opened 2021-02-17
Fix unprivileged packet matching on Linux

### LINKED ISSUE #146: It hangs in unprivileged mode
ping.go example hangs forever but I see ICMP requests and replies:
$ ./ping -c 5 www.yandex.ru
PING www.yandex.ru (77.88.55.80):
^C
--- www.yandex.ru ping statistics ---
5 packets transmitted, 0 packets received, 0 duplicates, 100% packet loss
round-trip min/avg/max/stddev = 0s/0s/0s/0s

tcpdump output 5 requests and replies:
15:10:34.417077 IP 192.168.0.3 > 77.88.55.80: ICMP echo request, id 29, seq 0, length 24
15:10:34.423019 IP 77.88.55.80 > 192.168.0.3: ICMP echo reply, id 29, seq 0, length 24
....

$ getcap ./ping
./ping = cap_net_raw+ep
$ sysctl net.ipv4.ping_group_range
net.ipv4.ping_group_range = 0	2147483647
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal
$ go version
go version go1.15.8 linux/amd64

Should it send UDP packets instead of ICMP in unprivileged mode?
It works in privileged mode as expected:
$ ./ping --privileged www.yandex.ru
PING www.yandex.ru (5.255.255.5):
24 bytes from 5.255.255.5: icmp_seq=0 time=8.114374ms ttl=55
24 bytes from 5.255.255.5: icmp_seq=1 time=8.309559ms ttl=55

Any ideas please?

CHTJonas:
62f79f1f4f36fc821eebb131548270d4d8b68653 changed the matching of ICMP IDs so that it happened in both privileged an unprivileged modes. I played around and tested this on macOS and Windows but not on Linux which, it seems, behaves differently.

This introduces a platform-dependant matching function using build constraints. Fixes #146.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -541,8 +541,7 @@ func (p *Pinger) processPacket(recv *packet) error {
 
 	switch pkt := m.Body.(type) {
 	case *icmp.Echo:
-		// Check if the reply has the ID we expect.
-		if pkt.ID != p.id {
+		if !p.matchID(pkt.ID) {
 			return nil
 		}
 
--- a/utils_linux.go
+++ b/utils_linux.go
@@ -1,4 +1,4 @@
-// +build !linux,!windows
+// +build linux
 
 package ping
 
@@ -9,8 +9,11 @@ func (p *Pinger) getMessageLength() int {
 
 // Attempts to match the ID of an ICMP packet.
 func (p *Pinger) matchID(ID int) bool {
-	if ID != p.id {
-		return false
+	// On Linux we can only match ID if we are privileged.
+	if p.protocol == "icmp" {
+		if ID != p.id {
+			return false
+		}
 	}
 	return true
 }
--- a/utils_windows.go
+++ b/utils_windows.go
@@ -14,3 +14,11 @@ func (p *Pinger) getMessageLength() int {
 	}
 	return p.Size + 8 + ipv6.HeaderLen
 }
+
+// Attempts to match the ID of an ICMP packet.
+func (p *Pinger) matchID(ID int) bool {
+	if ID != p.id {
+		return false
+	}
+	return true
+}


CHTJonas:
Looks like CI is failing because we're using the latest version of goreleaser which is trying to build for darwin_arm64 (Apple Silicon) but we're using Go version 1.15 which doesn't support that OS/arch combo. I'll open another PR to bump the Go version.

jraby:
I was trying to figure out why go-ping hung in unpriviliged / datagram mode and ended up here.

It seems to reason why message ID are not comparable / ignored on linux  is because the kernel uses the socket's source port as the ID for outgoing message.
The datagram socket is created with a 0 port which tells the kernel to choose a free port.
The port number is then used to demultiplex incoming icmp replies and send them to the corresponding socket.

Anyways, lots of words to say that ignoring the ID in this case is quite correct.

---

Ref: https://lwn.net/Articles/443051/

```
[pid 3322448] socket(AF_INET, SOCK_DGRAM, IPPROTO_ICMP) = 3
[pid 3322448] bind(3, {sa_family=AF_INET, sin_port=htons(0), sin_addr=inet_addr("0.0.0.0")}, 16) = 0
[pid 3322448] fcntl(3, F_GETFL)         = 0x2 (flags O_RDWR)
[pid 3322448] fcntl(3, F_DUPFD_CLOEXEC, 0) = 7
[pid 3322448] fcntl(7, F_GETFL)         = 0x2 (flags O_RDWR)
[pid 3322448] fcntl(7, F_SETFL, O_RDWR|O_NONBLOCK) = 0
[pid 3322448] getsockopt(7, SOL_SOCKET, SO_TYPE, [2], [4]) = 0
[pid 3322448] getsockname(7, {sa_family=AF_INET, sin_port=htons(34), sin_addr=inet_addr("0.0.0.0")}, [112->16]) = 0

----

21:35:36.244849 wlan0 Out IP 172.16.1.156 > 172.217.13.196: ICMP echo request, id 34, seq 0, length 24
21:35:36.256708 wlan0 In  IP 172.217.13.196 > 172.16.1.156: ICMP echo reply, id 34, seq 0, length 24

```

CHTJonas:
@jraby thanks for the confirmation I wasn't barking up the wrong tree - good detective work! Closing as this has been merged into `master` in 5f9dd90.

PR #148, opened 2021-02-17
Bump Go version to v1.16

CHTJonas:
Updates the version of Go used in Circle to 1.16 so that we can use goreleaser to test building for the darwin_arm64 target.

Discovered via #147.


SuperQ (approved):

CHTJonas:
No longer seems necessary. I suspect goreleaser fixed at their end.

PR #150, opened 2021-03-10
Compute all statistics as packets are received 

jraby:
This PR implements [Welford's online algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) to compute the standard deviation of the round trip times as packets are received without storing the full list. (the average is taken from this algorithm too)
As for min and max, they are also computed as we go.

I didn't touch the `rtts` array since it is part of the public interface of the Statistics, but with this PR it is not needed to get the stats.

I wasn't sure what to do about the tests which set `p.PacketsRecv` and then filled `p.rtts` with test values.
Since the packet count is an inherent part of the algorithm, I opted to move the increment into the `updateStatistic` method and repeatedly call it with test data.

One other thing: I think calling `Statistics()` from another goroutine could yield incorrect results since `updateStatistics()` could be running at the same time. Adding a RWLock should fix that. (I haven't done it yet since I do not use Pinger in this fashion and this particular problem was present in the original code where rtts can be modified while `Statistics()` is called)

There's also a commit that renames `outPkt` to `inPkt` in `processPacket` since that function is dealing with received packets.

(And this is all branched off b92053f because otherwise go-ping doesn't work on linux at the moment)

It might be easier to review commit by commit since they are self contained.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -541,8 +541,7 @@ func (p *Pinger) processPacket(recv *packet) error {
 
 	switch pkt := m.Body.(type) {
 	case *icmp.Echo:
-		// Check if the reply has the ID we expect.
-		if pkt.ID != p.id {
+		if !p.matchID(pkt.ID) {
 			return nil
 		}
 
--- a/ping.go
+++ b/ping.go
@@ -532,7 +532,7 @@ func (p *Pinger) processPacket(recv *packet) error {
 		return nil
 	}
 
-	outPkt := &Packet{
+	inPkt := &Packet{
 		Nbytes: recv.nbytes,
 		IPAddr: p.ipaddr,
 		Addr:   p.addr,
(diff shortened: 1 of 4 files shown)


ping.go
```diff
@@ -137,6 +137,13 @@ type Pinger struct {
 	// Number of duplicate packets received
 	PacketsRecvDuplicates int
 
+	// Round trip time statistics
+	MinRtt    time.Duration
+	MaxRtt    time.Duration
+	AvgRtt    time.Duration
+	StdDevRtt time.Duration
```
SuperQ, line 144: Should these remain unexported? I don't think it's intended that these be set directly, only accessed via `Statistics()`. Otherwise there could be a data race if I'm understanding this correctly.

SuperQ:
Thanks, this is great.

Yes, we do want to keep `Pinger.rtts`, as some people use this library to get all of them for later processing (for example, calculating quantiles). But adding the live stats update makes it so we don't need it to calculate simple stats.

ping.go
jraby: You're right, I'll update the PR.

CHTJonas (approved):
LGTM - thanks for this! Happy with everything here and for the unprivileged matching to be merged in at the same time as this. I'll close #147 when I merge.

PR #151, opened 2021-03-10
Protect stats with RWMutex 

jraby:
This is more of a request for comments than a proper PR (and it is branched off #150  so only the last commit should be reviewed )

A RWMutex is added to protect access to  the Pinger's internal statistics.
Updates to `rtts` is moved to `updateStats` so they happen under the lock.

There's (at least) one issue however: PacketsSent and PacketsRecvDuplicates can still be updated while
Statistics() is running, and thus could yield strange results (sent<recv)

To fix this, atomics could be used, but it would require changing the types of the fields from int to int64, thus changing the public api of Pinger. (One could argue that those fields should probably not be exported to begin with, but that would break the api too)

Any thoughts?

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -143,6 +143,7 @@ type Pinger struct {
 	avgRtt    time.Duration
 	stdDevRtt time.Duration
 	stddevm2  time.Duration
+	statsmu   sync.RWMutex
 
 	// If true, keep a record of rtts of all received packets.
 	// Set to false to avoid memory bloat for long running pings.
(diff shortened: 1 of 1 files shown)


SuperQ:
Yessssssssssssssssssss.

jraby:
Revisiting this, I think it could be possible to add a unexported uint64 (or int64) for packetsSent and duplicatesrecv.
Updatng them using `atomic.AddUint64` and accessing them in the stats with `atomic.LoadUint64`.

The exported fields can stay there and be updated as they are right now, but they wouldn't be used in Statistics().
With that, I think `Statistics` would be safe, or at least a lot safer.

CHTJonas (approved):
Nice one! It's a 👍 from me for a squash-and-merge.

jraby:
Rebased

ping.go
```diff
@@ -143,6 +143,7 @@ type Pinger struct {
 	avgRtt    time.Duration
 	stdDevRtt time.Duration
 	stddevm2  time.Duration
+	statsmu   sync.RWMutex
```
SuperQ, line 146: Minor nit about Go naming style, this should be CamelCase. How about`statusMtx`?

jraby: `statsMu`  ?

Most examples I see use `Mu`, not `Mtx`

SuperQ: `Mu` is fine too.

SuperQ (approved):
LGTM

Issue #142, opened 2021-02-08
Wrap sequence numbers


SuperQ:
ICMP sequence numbers are only 16 bits, so we need to wrap the counter when we hit 65535.

See: https://github.com/SuperQ/smokeping_prober/issues/51

CHTJonas:
Go's uint16 should wrap as expected (https://play.golang.org/p/mRbqYpALimF) so I think we just need to change the datatype from int to that:

https://github.com/go-ping/ping/blob/ab39f29b51f8737862e1b692f31fbcc4ba87d290/ping.go#L176

Does this raise concerns about the inflight tracking in the `awaitingSequences` map? My own view is that if someone sends 65536 packets in the space of a RTT timeout then that's their own fault.

SuperQ:
Yes, it's going to possibly need some extra checks around the awaitingSequences. :-(

Another way to handle this is to reset the sequence number and generate a new tracking ID.

awaitingSequences can be a `map[trackingId][int]struct{}{}`. Then we'll have to make the ID a `[]trackingId`.

Another good reason to use UUID for the tracking id.

SuperQ:
Maybe @jraby wants to work on this one. :smile:

PR #152, opened 2021-03-16
Use a UUID for packet tracking

SuperQ:
Replace the random 8 byte tracker with a 16 byte UUID.
* Implement as a slice so that it can be extended later.
* Deprecate the exported `Pinger.Tracker` value.
* Wrap sequence numbers into new UUID packet tracking.

Fixes: https://github.com/go-ping/ping/issues/142


CHTJonas (approved):
LGTM!

(This makes the minimum `p. Size` 24 which I'm fine with. It's possible to start truncating the tracker and timestamp if the user-requested size is small, like `/bin/ping` does, however this is something for the future.)

SuperQ:
It looks like the code will silently ignore `Pinger.Size` smaller than the minimum.

SuperQ:
CC @mem, what do you think of this?

ping.go
```diff
@@ -633,7 +641,11 @@ func (p *Pinger) sendICMP(conn *icmp.PacketConn) error {
 		dst = &net.UDPAddr{IP: p.ipaddr.IP, Zone: p.ipaddr.Zone}
 	}
 
-	t := append(timeToBytes(time.Now()), uintToBytes(p.Tracker)...)
+	currentUUID, err := p.trackerUUIDs[len(p.trackerUUIDs)-1].MarshalBinary()
+	if err != nil {
+		return fmt.Errorf("unable to marshal UUID binary: %w", err)
+	}
+	t := append(timeToBytes(time.Now()), currentUUID...)
```
mem, line 648: this bit of logic is reproduced in multiple locations. Should it be moved to a separate function, so that it's easier to extend in the future?

ping.go
```diff
@@ -66,14 +65,15 @@ import (
...
+	"github.com/google/uuid"
 	"golang.org/x/net/icmp"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
 )
 
 const (
 	timeSliceLength  = 8
-	trackerLength    = 8
+	trackerLength    = 16
```
mem, line 76: maybe:

```suggestion
	trackerLength    = len(uuid.UUID{})
```

as the expression ` len(uuid.UUID{})` is a compile-time constant.

mem (commented):
LGTM, just a couple of minor comments

ping.go
SuperQ: Good idea.

ping.go
SuperQ: Hmm, possibly. Most of the duplication is in the tests. I'm also planning to make changes after this PR to start implementing the tracker + sequence number wrapping.

I'd say let's leave it for later refactoring.

mem: cool, thanks!

ping.go
```diff
@@ -587,6 +597,27 @@ func (p *Pinger) recvICMP(
 	}
 }
 
+// getPacketUUID scans the the tracking slice for matches.
```
CHTJonas, line 600:
```suggestion
// getPacketUUID scans the tracking slice for matches.
```

cmd/ping/ping.go
```diff
@@ -93,6 +93,6 @@ func main() {
 	fmt.Printf("PING %s (%s):\n", pinger.Addr(), pinger.IPAddr())
 	err = pinger.Run()
 	if err != nil {
-		fmt.Printf("Failed to ping target host: %s", err)
+		fmt.Printf("Failed to ping target host: %s\n", err)
```
CHTJonas, line 96:
```suggestion
		fmt.Println("Failed to ping target host:", err)
```

ping.go
```diff
@@ -385,6 +391,10 @@ func (p *Pinger) Run() error {
 		logger = NoopLogger{}
 	}
 
+	if p.Size < timeSliceLength+trackerLength {
+		return fmt.Errorf("Size %d is less than minimum required size %d", p.Size, timeSliceLength+trackerLength)
```
CHTJonas, line 395:
```suggestion
		return fmt.Errorf("size %d is less than minimum required size %d", p.Size, timeSliceLength+trackerLength)
```

CHTJonas (changes requested):
Nice one! And sorry it took so long getting round to give this a second look... a few nits but otherwise looks good.

SuperQ:
@CHTJonas Fixed up.

CHTJonas (approved):

CHTJonas (changes requested):
Re-re-rebase for the Nth time! 😛

CHTJonas (approved):
LGTM! 🚀

mem (approved):
LGTM!

CHTJonas:
Closed by #177 in 38182647687148eb7c963dc57dc62456b12aa0ae.

PR #153, opened 2021-03-20
Wrap errors when using fmt.Errorf

CHTJonas:
Nit: wrap errors using `%w` to allow the use of the `errors.Is` function. https://blog.golang.org/go1.13-errors

Currently the returned error is only used internally and it simply gets logged if not `nil`:

https://github.com/go-ping/ping/blob/22d47451d074c67c29d130af1ce96f45edb4e30d/ping.go#L399

However #16 wants to change this and return errors to the end user so this PR will invariably help when #16 is resolved.


SuperQ (approved):
LGTM

sparrc (approved):

PR #154, opened 2021-03-24
Allow intervals of 0 seconds

mem:
time.NewTicker doesn't like an interval of 0, but it's useful in order
to send a "flood ping" (see for example
https://www.freebsd.org/cgi/man.cgi?ping(8) or
https://linux.die.net/man/8/ping). It would be up to the user of the
library to limit the number of in-flight pings in the manner described
in the manpages.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -407,8 +407,18 @@ func (p *Pinger) Run() error {
 
 	timeout := time.NewTicker(p.Timeout)
 	defer timeout.Stop()
-	interval := time.NewTicker(p.Interval)
-	defer interval.Stop()
+
+	var intervalCh <-chan time.Time
+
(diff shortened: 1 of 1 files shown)


CHTJonas (changes requested):
This works but I wonder if a simpler approach would just be to set `p.Interval = time.Nanosecond`?

mem:
> This works but I wonder if a simpler approach would just be to set `p.Interval = time.Nanosecond`?

Sure, that works. That was in fact my first approach. It has the nasty side-effect of the ticker triggering as fast as it can and it keeps firing until all the packets have been received. This implementation does the same with a little less overhead and blocking that select interval case as soon as all the required packets have been sent out.

PR #155, opened 2021-03-24
Add OnSetup callback

mem:
It allows users to do something after the listening socket is set up but
before the first packet is sent. One specific use case is measuring the
time it takes the system to *send* the first packet.


SuperQ (approved):
LGTM

CHTJonas (approved):
LGTM

Issue #157, opened 2021-03-25
[Threading] Crash ping.(*Pinger).recvICMP


STiAT:
Hi there,

not sure if it's on the usage or library side, but I create a new pinger in a thread calling pinger like this:
```
func ping_host(host string, count int, timeout int) error {
	pinger, err := ping.NewPinger(host)
	if err != nil {
		return err
	}

	if runtime.GOOS == "windows" {
		// Windows requires SetPrivileged
		pinger.SetPrivileged(true)
	} else {
		// others can perfectly well do that with a system setting
		// either
		// setcap cap_net_raw=+ep /opt/slacheck/slacheck
		// or
		// sudo sysctl -w net.ipv4.ping_group_range="0 2147483647"

		pinger.SetPrivileged(false)
	}

	pinger.Timeout = time.Second * time.Duration(timeout*count)
	pinger.Interval = time.Second * time.Duration(timeout)
	pinger.Count = count

	err = pinger.Run()

	if err != nil {
		return err
	}

	stats := pinger.Statistics()
	if stats.PacketLoss == 100 {
		return errors.New("Host Timeout: " + host)
	}

	return nil
}
```

Once in a while (every 400-600 hosts I ping) I get the following crash:
```
panic: close of closed channel

goroutine 3301 [running]:
github.com/go-ping/ping.(*Pinger).recvICMP(0xc00005f540, 0xc001b02260, 0xc0011286c0, 0xc0018ac7e0, 0x0, 0x0)
        C:/Users/ngrgeo/devel/go/pkg/mod/github.com/go-ping/ping@v0.0.0-20210312085107-d90f3778a8a3/ping.go:518 +0x372
created by github.com/go-ping/ping.(*Pinger).Run
        C:/Users/ngrgeo/devel/go/pkg/mod/github.com/go-ping/ping@v0.0.0-20210312085107-d90f3778a8a3/ping.go:401 +0x299
```

This may be a bug in the threading / mutexes, seems recvICMP is not really thread safe? That said, I'm running 100 threads in parallel (have to ping a few thousand hosts).

I didn't take a closer look into the implementation yet, but I thought I'll raise the issue - can still work on that later.

//G

STiAT:
Oh wow, that was fast. Will test that on Monday. Thanks!

PR #159, opened 2021-03-25
Reorder channel selection order and make Stop() idempotent

CHTJonas:
This PR:

- Reorders the channels in the main select clause of the `Run()` function so that we always deal with received packets first rather than sending new packet out. This is so that the TTL values don't go stale when pinging very rapidly (cf. #154).
- Improves CPU usage slightly by greedily calling `Stop()` on tickers where possible.
- Removes the initial packet sent around line 403 which I can't really see a need for.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -409,19 +409,18 @@ func (p *Pinger) Run() error {
 
 	timeout := time.NewTicker(p.Timeout)
 	interval := time.NewTicker(p.Interval)
+	defer func() {
+		interval.Stop()
+		timeout.Stop()
+		wg.Wait()
+	}()
 
 	for {
 		select {
 		case <-p.done:
-			interval.Stop()
-			timeout.Stop()
-			wg.Wait()
 			return nil
 		case <-timeout.C:
 			close(p.done)
-			interval.Stop()
-			timeout.Stop()
-			wg.Wait()
 			return nil
 		case r := <-recv:
 			err := p.processPacket(r)
@@ -442,9 +441,6 @@ func (p *Pinger) Run() error {
 		}
 		if p.Count > 0 && p.PacketsRecv >= p.Count {
 			close(p.done)
-			interval.Stop()
-			timeout.Stop()
-			wg.Wait()
 			return nil
 		}
 	}
--- a/ping.go
+++ b/ping.go
@@ -417,6 +417,11 @@ func (p *Pinger) Run() error {
 		wg.Wait()
 	}()
 
+	err = p.sendICMP(conn)
+	if err != nil {
+		return err
+	}
+
 	for {
 		select {
 		case <-p.done:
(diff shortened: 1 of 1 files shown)


SuperQ:
I think I prefer to stick to the defer Stop() method. Having to maintain stops on all function exits is fragile.

CHTJonas:
How about a single function which makes all the necessary `Stop()` calls? I could factor this out into a separate PR if necessary (and maybe merge with parts of #116 minus the error bit)?

SuperQ:
It's not the number of function calls, it's the fact that you need to make sure they're all there at any exit from the function. This is why defer is used.

CHTJonas:
Updated to use the defer-style syntax which I think looks better. It's objectively less code to worry about/forget in that select clause. I've also included the idempotent `pinger.Stop()` because it made sense to.

SuperQ (approved):
LGTM, nice.

ping.go
```diff
@@ -407,51 +408,57 @@ func (p *Pinger) Run() error {
 		handler()
 	}
 
-	err = p.sendICMP(conn)
-	if err != nil {
-		return err
-	}
```
mem, line 413: this is necessary so that the first ping is sent when Run is called and not a p.Interval time later. The first time a ticker ticks is after the interval is elapsed, not at 0, so the code on current line 429 runs for the first time after that time. This send is here so that the first ping is sent at t=0, so to speak.

mem (commented):
Adding a comment as to why the first p.sendICMP(conn) is necessary.

ping.go
CHTJonas: Yeah I realised this sometime afterwards while looking over my PR again. I kind of decided I didn't care enough to change it back but I'm happy to do so.

CHTJonas:
This also fixes #157.

ping.go
```diff
@@ -407,51 +408,62 @@ func (p *Pinger) Run() error {
...
 }
 
 func (p *Pinger) Stop() {
-	close(p.done)
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	open := true
+	select {
+	case _, open = <-p.done:
```
sparrc, line 460: the p.done channel has been changed to `chan interface{}` right? I actually didn't think that this would compile and Im not entirely sure I understand the intended behavior. You are setting a boolean variable to `interface{}`? Does that make it `false`?

sparrc: nevermind, `_` is the interface{}, i see now

sparrc (approved):
lgtm, just to double-verify this bullet point in the original description does not apply anymore correct? 

> - Removes the initial packet sent around line 403 which I can't really see a need for.

IIUC the initial packet send was re-added by https://github.com/go-ping/ping/pull/159/commits/91995fe2b7830c3ab95c9e54141be680c22348a0

CHTJonas:
Yep correct on all accounts @sparrc!

chenk008:
In my opinion, there is no guarantee order in `select` clause.

PR #160, opened 2021-03-27
Log errors returned by recvICMP

CHTJonas:
Currently we throw errors from the `pinger.recvICMP` function away which is very much non-ideal. While having a library print logs to the stdout is also far from ideal, it's arguably a less evil than not reporting them _at all_ at least until such a time that we can implement a proper error-returning or logging interface (#16).

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -415,8 +415,13 @@ func (p *Pinger) Run() error {
 	recv := make(chan *packet, 5)
 	defer close(recv)
 	wg.Add(1)
-	//nolint:errcheck
-	go p.recvICMP(conn, recv, &wg)
+
+	go func() {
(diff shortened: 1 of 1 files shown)


SuperQ (approved):
LGTM

mem (approved):

SuperQ:
The other way we could handle this is to use an [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) instead of a simple waitgroup.

CHTJonas:
Rebased on top of the logger changes.

mem (approved):

CHTJonas:
Ping @sparrc for a review please! 🥺

CHTJonas:
Superseded by ff8be3320020b96856d3d48505bcf8926809752e.

PR #161, opened 2021-04-01
Add Logger interface to avoid logging directly

### LINKED ISSUE #16: Library should return errors and not printf messages

mem:
This is a simple level-based log interface that allows users of the
library to control and customize the output for example by providing an
standard logger with different settings (timestamps, prefixes, etc).

Closes #16 

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,0 +1,53 @@
+package ping
+
+import "log"
+
+type Logger interface {
+	Fatalf(format string, v ...interface{})
+	Errorf(format string, v ...interface{})
+	Warnf(format string, v ...interface{})
+	Infof(format string, v ...interface{})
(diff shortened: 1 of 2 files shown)


mem:
I was looking at #103 and since I cannot update that PR, I uploaded this one. It's slightly different but the goal is the same.

SuperQ (approved):
LGTM, Thanks!

CHTJonas (approved):
👍🏻🚀

PR #162, opened 2021-04-02
Use exponential backoff for read deadline

mem:
Before this change, pinging any target once takes ~ 100 ms (the time it
takes for Run() to return, not the RTT). After this change, it takes a
time comparable to the RTT. The change comes from ReadFrom blocking for
the specified delay if the done signal hasn't fired yet. It also
improves a little the time it takes to ping a single target multiple
times.

Since this is exponential backoff in the way Ethernet does it (random
delay, with an increasing maximum value, only changing in case of a
timeout), the delay neither necessarily doubles nor increases each time
the operation has to be retried, but it does eventually settle on the
higher end of the possibilities if there are many timeouts.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -537,8 +537,8 @@ func (p *Pinger) recvICMP(
 ) error {
 	defer wg.Done()
 
-	// Start by waiting for 100 µs and increase to a possible maximum of ~ 100 ms.
-	expBackoff := newExpBackoff(100*time.Microsecond, 10)
+	// Start by waiting for 50 µs and increase to a possible maximum of ~ 100 ms.
+	expBackoff := newExpBackoff(50*time.Microsecond, 11)
 	delay := expBackoff.Get()
 
 	for {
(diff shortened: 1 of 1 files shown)


ping.go
```diff
@@ -498,19 +498,42 @@ func (p *Pinger) Statistics() *Statistics {
...
+}
+
 func (p *Pinger) recvICMP(
 	conn *icmp.PacketConn,
 	recv chan<- *packet,
 	wg *sync.WaitGroup,
 ) error {
 	defer wg.Done()
+
+	// Start by waiting for 100 µs and increase to a possible maximum of ~ 100 ms.
```
SuperQ, line 526: Pings over virtual interfaces, say in a local VM network, can be sub 100µs. Maybe start this at 50µs? Not sure what you were basing this value on.

mem: sounds, good, changed.

SuperQ (approved):
LGTM

CHTJonas (approved):

PR #158, opened 2021-03-25
Don't fragment

mem:
This PR adds a "don't fragment" option that sets the IPv4 don't fragment flag so that ping fails if the packet exceeds the MTU for the link.

PR #166, opened 2021-04-29
Provide an abstraction over icmp.PacketConn

mem:
The differences between ipv4 and ipv6 APIs can be moved to a single type
so that we don't need to keep track of them all over the code.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -69,6 +69,7 @@ import (
 	"golang.org/x/net/icmp"
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


mem:
This is extracted from #158 in the hopes that it makes it easier to review. I had to modify the Run method to make it testable, and that's what making the change noisy, but I didn't want this to be intermingled with the other review.

ping.go
```diff
@@ -380,89 +380,93 @@ func (p *Pinger) SetLogger(logger Logger) {
...
-			err := p.processPacket(r)
-			if err != nil {
-				// FIXME: this logs as FATAL but continues
-				logger.Fatalf("processing received packet: %s", err)
-			}
-		case <-interval.C:
-			if p.Count > 0 && p.PacketsSent >= p.Count {
-				interval.Stop()
-				continue
+	g.Go(func() error {
```
SuperQ, line 424: Is there a good way to extract out this anon func? I find these large inline funcs a bit ugly.

SuperQ (approved):
Minor nit about large inline func, but otherwise LGTM

ping.go
mem: OK, I gave this a go. Naming is ... hmm...

SuperQ:
Poke @CHTJonas

CHTJonas (approved):
LGTM 🚀

PR #171, opened 2021-07-12
added SetSrcAddress and GetIfaceIpv4Addr methods

shreyaskapale:
added **SetSrcAddress** and **GetIfaceIpv4Addr** methods

1. **SetSrcAddress** takes in either interface name of the source example "eth0" or the address "192.168.1.239" and sets source ip in the Pinger Class.
2. **GetIfaceIpv4Addr** is a helper function which returns a ipv4 address when interface name is supplied.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -61,6 +61,7 @@ import (
 	"math"
 	"math/rand"
 	"net"
+	"regexp"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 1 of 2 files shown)


ping.go
```diff
@@ -339,6 +340,46 @@ func (p *Pinger) Addr() string {
 	return p.addr
 }
 
+// SetSrcAddr set source by interface name or ipv4 address
+// Interface name can be "eth0" or ip like "192.168.1.239"
+func (p *Pinger) SetSrcAddr(iface string) error {
+
+	ipPattern := regexp.MustCompile(`(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}`)
+	if ipPattern.MatchString(iface) {
```
SuperQ, line 348: This should probably use go [`net.ParseIP()`](https://golang.org/pkg/net/#ParseIP).

SuperQ:
This needs to handle both IPv4 and Ipv6. This should also include tests.

ping.go
shreyaskapale: Cool, I will replace it with net.ParseIP().

shreyaskapale:
> 
> 
> This needs to handle both IPv4 and Ipv6. This should also include tests.

I will add try adding ipv6 handler and include tests for the same.

shreyaskapale:
> This needs to handle both IPv4 and Ipv6. This should also include tests.
Please review  the new commits I have added handlers for both and also included the tests for the same

ping.go
```diff
@@ -339,6 +339,58 @@ func (p *Pinger) Addr() string {
...
+		return err
+	} else {
+		p.Source = ipAddr[0].String() // As the private IP range is generally in ipv4, so as default it sets ipv4 of the iface
+		return nil
+	}
+}
+
+// GetIfaceAddrs is helper function which returns ipv4 and ipv6 as a array of net.IP type of supplied interface name
+// GetIfaceAddrs returns [ipv4,ipv6] of type net.IP which can be parsed to string
+func (p *Pinger) GetIfaceAddrs(interfaceName string) (addrs []net.IP, err error) {
```
CHTJonas, line 364: This is a helper function so let's make it private i.e. start with a lowercase letter.

ping.go
```diff
@@ -339,6 +339,58 @@ func (p *Pinger) Addr() string {
 	return p.addr
 }
 
+// SetSrcAddr set source by interface name or ipv4 address
+// Interface name can be "eth0" or ip like "192.168.1.239"
```
CHTJonas, line 343: The API feel confused here, imo. I think we should separate setting the source address to an IP literal vs. setting it to an address that's attached to a given interface.

ping.go
```diff
@@ -339,6 +339,58 @@ func (p *Pinger) Addr() string {
...
+		} else {
+			p.SetNetwork("ipv6")
+		}
+		p.Source = iface
+		return nil
+	}
+	if ipAddr, err := p.GetIfaceAddrs(iface); err != nil {
+		return err
+	} else {
+		p.Source = ipAddr[0].String() // As the private IP range is generally in ipv4, so as default it sets ipv4 of the iface
```
CHTJonas, line 357: `ipAddr` is an array but we always take the first value - this feels like a likely source of subtle bugs for end users. Not saying it isn't the least-worst thing to do, but maybe we should document. As an aside, the comment is a bit misleading here - there's no guarantee this IP will be a private one, especially if it's a v6 address.

CHTJonas (changes requested):

ping.go
shreyaskapale: Cool, I will make the changes and commit it. Thanks for the review 🎉

ping.go
shreyaskapale: I think we can keep this as public, it can be used by the end user if needed to fetch IPv4 and IPv6 via interface.

ping.go
shreyaskapale: Are you suggesting to make 2 separate functions like, setSrcByIP and setScrByIface ?

CHTJonas: Unless there's a really compelling reason to combine them, I feel like 2 separate function is more intuitive, yes.

SuperQ:

PR #176, opened 2021-08-27
Add packet timeout callback #50

hamptonmoore:
My approach to fixing #50. It should be noted that CheckAwaitingSequences will only be called everytime a ping is attempted so accuracy in terms of PacketTimeout is as accurate as Interval.


SuperQ:
This needs a rebase now that #177 is merged.

SuperQ:
Another way to improve the timeout reporting accuracy would be to setup a [time.Timer](https://pkg.go.dev/time#Timer) in the main packet loop.

SuperQ:
This needs a rebase.

SuperQ:

PR #184, opened 2021-09-14
Maintenance: add more linters to golangci-lint

qdm12:
- Use a maphash source instead of time based global variable
- Add exported wrapping errors
- Add unexported wrapping errors to add more context to the errors
- Small code changes to satisfy all linters (e.g. adding inlined constants)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,4 @@
 linters-settings:
-  maligned:
-    suggest-new: true
   misspell:
     locale: US
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,6 @@ issues:
         - goerr113
 
 linters:
-  disable-all: true
   enable:
     - asciicheck
     - bodyclose
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,11 +16,9 @@ linters:
     - asciicheck
     - bodyclose
     # - cyclop - enable later
-    - deadcode
     - dogsled
     - dupl
     - durationcheck
-    - errcheck
     - errname
     - errorlint
     - exhaustive
@@ -41,11 +39,8 @@ linters:
     - gomoddirectives
     - goprintffuncname
     - gosec
-    - gosimple
-    - govet
     - importas
     - importas
-    - ineffassign
     - lll
     - makezero
     - misspell
@@ -61,15 +56,10 @@ linters:
     - revive
     - rowserrcheck
     - sqlclosecheck
-    - staticcheck
-    - structcheck
     - thelper
     - tparallel
-    - typecheck
     - unconvert
     - unparam
-    - unused
-    - varcheck
     - wastedassign
     - whitespace
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,79 @@
----
+linters-settings:
+  maligned:
+    suggest-new: true
+  misspell:
+    locale: US
+
 issues:
   exclude-rules:
-  - path: _test.go
-    linters:
+    - path: _test\.go
+      linters:
+        - dupl
+        - goerr113
+
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
+    - bodyclose
+    # - cyclop - enable later
+    - deadcode
+    - dogsled
+    - dupl
+    - durationcheck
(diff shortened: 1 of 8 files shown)


ping.go
```diff
@@ -624,7 +637,7 @@ func (p *Pinger) processPacket(recv *packet) error {
 	var m *icmp.Message
 	var err error
 	if m, err = icmp.ParseMessage(proto, recv.bytes); err != nil {
-		return fmt.Errorf("error parsing icmp message: %w", err)
+		return fmt.Errorf("%w: %s", errParseICMPMessage, err)
```
SuperQ, line 640: This should probably be `"%w: %w"`.

.golangci.yml
```diff
@@ -1,6 +1,79 @@
...
+    - nolintlint
+    - prealloc
+    - predeclared
+    - predeclared
+    - promlinter
+    - revive
+    - rowserrcheck
+    - sqlclosecheck
+    - staticcheck
+    - structcheck
```
SuperQ, line 66: This [linter is deprecated](https://github.com/golangci/golangci-lint/issues/1841), please remove it.

.golangci.yml
```diff
@@ -1,6 +1,79 @@
...
+    - thelper
+    - tparallel
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
```
SuperQ, line 73: This [linter is deprecated](https://github.com/golangci/golangci-lint/issues/1841), please remove it.

SuperQ (changes requested):
While I like some of the changes, I very much dislike manually managing the list of enabled linters. This is going to lead to toil.

Please do not `disable-all`, leave the recommended default enabled linters as they are.

If you want to enable specific linters, go ahead. But don't leave us with a bunch of tech debt of manually managing this.

ping.go
qdm12: You can only wrap with a single error. If you try with 2:

> fmt.Errorf call has more than one error-wrapping directive %wprintf

The idea is that functions calling this unexported function can check what block of code failed by using `errors.Is(err, errParseICMPMessage)`, without relying on errors defined in the `icmp` package (if there is any 🤔). We keep the original error as a string for the user to see in logs etc.

.golangci.yml
```diff
@@ -1,6 +1,79 @@
...
+        - dupl
+        - goerr113
+
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
+    - bodyclose
+    # - cyclop - enable later
+    - deadcode
```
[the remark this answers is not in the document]
qdm12: `deadcode` is enabled by default (see [here](https://golangci-lint.run/usage/linters/)). If I remove `disable-all`, it will always be enabled then 😄 Instructions unclear 🤖 

As are `structcheck`, `unused` and `varcheck` too.

qdm12:
@SuperQ I would suggest to trust `golangci-lint` and not `disable-all` and leave all of the default enabled, even if they are 'deprecated'. Golangci-lint will eventually remove them in an upcoming release.

I'll also remove the linters enabled by default (for now) which I manually listed so they get automatically removed when golangci-lint gets upgraded to a release removing deprecated ones.

qdm12:
Also I made https://github.com/qdm12/ping/pull/1 to save a few bytes for structures (the order of fields matter in Go), let me know if you want to merge that in this branch or just discard it, thanks!

ping.go
SuperQ: Makes sense.

.golangci.yml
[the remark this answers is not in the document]
SuperQ: Let's explicitly disable the deprecated linters.

Something like this:

```yaml
linters:
  enable:
    - revive
  disable:
    # Disable soon to deprecated[1] linters that lead to false
    # positives when build tags disable certain files[2]
    # 1: https://github.com/golangci/golangci-lint/issues/1841
    # 2: https://github.com/prometheus/node_exporter/issues/1545
    - deadcode
    - unused
    - structcheck
    - varcheck
```

(This is what I do in prometheus/node_exporter)

SuperQ:
This needs a rebase.

SuperQ:

PR #185, opened 2021-09-14
Feat: `Pinger`'s `RunWithContext` method

qdm12:
- [x] Method with comment
- [x] Unit test
  - [ ] Not sure how to go around testing this without root in the CI, any idea?
- [x] This is used in 'production' [here](https://github.com/qdm12/gluetun/blob/f7e4331e933a396c9279b48340e1284a98fe50d5/internal/healthcheck/health.go#L61-L77)
- [ ] Document in readme?

### DIFF
--- a/ping_test.go
+++ b/ping_test.go
@@ -795,7 +795,6 @@ func TestRunWithContext(t *testing.T) {
 			t.Parallel()
 
 			pinger := New("127.0.0.1")
-			pinger.SetPrivileged(true)
 			pinger.Count = testCase.count
 
 			err := pinger.RunWithContext(testCase.ctx)
--- a/ping_test.go
+++ b/ping_test.go
@@ -2,6 +2,7 @@ package ping
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


SuperQ (approved):
LGTM, nice.

ping_test.go
```diff
@@ -762,3 +763,46 @@ func TestRunOK(t *testing.T) {
...
+		},
+	}
+
+	for name, testCase := range testCases {
+		testCase := testCase
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			pinger := New("127.0.0.1")
+			pinger.SetPrivileged(true)
```
SuperQ, line 798: Our CI environment can't test SetPrivileged packets.

CHTJonas (approved):
Great - thank you!

Unfortunately CircleCI doesn't support ICMP so we can't test using privileged pingers as we get `socket: operation not permitted`. Could you just remove the `pinger.SetPrivileged(true)` line so the tests can pass?

qdm12:
@SuperQ @CHTJonas I tried removing the `.SetPrivileged(true)` in https://github.com/go-ping/ping/pull/185/commits/5b70679101c4bc27ebe9fde4635ac342f1e98c6e

but the CI complains `ping_test.go:803: expected <nil> but got socket: permission denied`.

I believe you need to set as root `sysctl -w net.ipv4.ping_group_range="0 2147483647"` to fix that, so back to square one right? 🤔  Can we have root access anywhere on the CI perhaps before running the tests so we can run that command?

Alternatively, I can put that test in a separate test file with a build tag like:

```go
//go:build integration
// +build integration
```

So it would only run with `go test -tags integration` and not in the CI, but that's risky business too 😥

alexflint:
Hey all - I came looking for exactly the functionality in this PR. Any update on this?

qdm12:
@alexflint I guess you could just copy pasta that method as a function. That's what I did back then here: https://github.com/qdm12/gluetun/blob/7d824a51798f3e635e116dfece6056fd2ad4a2c6/internal/healthcheck/health.go#L59

tarndt:
Here is an alternate solution I use to get this functionality: 

```go
package netutil

import (
	"context"

	"github.com/go-ping/ping"
)

type Pinger struct {
	*ping.Pinger

	ctx       context.Context
	ctxCancel context.CancelFunc
}

func NewPinger(addr string) (*Pinger, error) {
	pinger, err := ping.NewPinger(addr)
	if err != nil {
		return nil, err
	}

	return &Pinger{Pinger: pinger}, nil
}

func (p *Pinger) Run(ctx context.Context) error {
	if err := ctx.Err(); err != nil {
		return err
	}

	p.ctx, p.ctxCancel = context.WithCancel(ctx)
	defer p.ctxCancel()

	go func() {
		<-ctx.Done()
		p.Pinger.Stop()
	}()

	return p.Pinger.Run()
}

func (p *Pinger) Stop() {
	p.ctxCancel()
}
```

qdm12:
I think we should close this, and leave it to the user to implement context aware methods. What do you think?

SuperQ:

PR #175, opened 2021-08-26
Add ability to set TTL

hamptonmoore:
Recently I needed the ability to set the Ping TTL for a project and couldn't find a package that would let me to do that so I modified this to let me. This is probably useful to other people aswell so I figure why not make a PR. Could probably do with a little bit of reformating how the TTL is sent from the Pinger to the icmpConn.

PR #186, opened 2021-10-08
Add ability to set TTL

cemremengu:
#175 rebased  (h/t @hamptonmoore)


CHTJonas:
We're blocked on CircleCI because it's not picking this up for some reason and "Required statuses must pass before merging"...

SuperQ:
Yea, I don't know why CircleCI isn't seeing this PR.

cemremengu:
@SuperQ should I just ignore the errors, log them as warning or fail fast by checking and returning the error?

what would be the most suitable approach?

PR #187, opened 2021-11-10
Always update Packet's addresses

gdetal:
Hi,

When a ping is performed on multicast addresses, the lib currently provides the address of the original destination in the callbacks. However, in this case the reply will come from another (existing) address. It is useful to provide the actual replied address to the callback instead of the original one (see the output below).

The output of ping without this patch is:

```
$ sudo ping --privileged ff02::1%ens3
PING ff02::1%ens3 (ff02::1%ens3):
24 bytes from ff02::1%ens3: icmp_seq=0 time=1.504454ms ttl=64
24 bytes from ff02::1%ens3: icmp_seq=0 time=1.553965ms ttl=64 (DUP!)
24 bytes from ff02::1%ens3: icmp_seq=0 time=1.580495ms ttl=64 (DUP!)
24 bytes from ff02::1%ens3: icmp_seq=0 time=1.596156ms ttl=64 (DUP!)
```

And with this patch:

```
$ sudo ./ping --privileged ff02::1%ens3
PING ff02::1%ens3 (ff02::1%ens3):
32 bytes from fe80::546f:f0ff:fe0e:a%ens3: icmp_seq=0 time=1.256761ms ttl=64
32 bytes from fe80::546f:f0ff:fe0e:29%ens3: icmp_seq=0 time=1.396183ms ttl=64 (DUP!)
32 bytes from fe80::546f:f0ff:fe0e:8%ens3: icmp_seq=0 time=1.447454ms ttl=64 (DUP!)
32 bytes from fe80::546f:f0ff:fe0e:e%ens3: icmp_seq=0 time=1.460754ms ttl=64 (DUP!)
```

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -211,6 +211,7 @@ type packet struct {
 	bytes  []byte
 	nbytes int
 	ttl    int
+	addr   *net.IPAddr
 }
 
 // Packet represents a received and processed ICMP echo packet.
@@ -574,7 +575,8 @@ func (p *Pinger) recvICMP(
 			}
 			var n, ttl int
 			var err error
-			n, ttl, _, err = conn.ReadFrom(bytes)
+			var addr net.Addr
+			n, ttl, addr, err = conn.ReadFrom(bytes)
 			if err != nil {
 				if neterr, ok := err.(*net.OpError); ok {
 					if neterr.Timeout() {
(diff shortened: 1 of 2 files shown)


SuperQ:

PR #190, opened 2021-11-29
Add configurable ICMP identifier

natesales:
This commit adds two functions to set and get a `Pinger`'s ICMP identifier field.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -232,6 +232,9 @@ type Packet struct {
 
 	// TTL is the Time To Live on the packet.
 	Ttl int
+
+	// ID is the ICMP identifier.
+	ID int
 }
 
 // Statistics represent the stats of a currently running or finished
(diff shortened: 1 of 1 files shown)


SuperQ (commented):
Good idea.

I wonder if we should also expose the `currentUUID` as well.

ping.go
```diff
@@ -232,6 +232,9 @@ type Packet struct {
 
 	// TTL is the Time To Live on the packet.
 	Ttl int
+
+	// Id is the ICMP identifier.
+	Id int
```
SuperQ, line 237:
```suggestion
	ID int
```

ping.go
```diff
@@ -385,6 +388,16 @@ func (p *Pinger) SetLogger(logger Logger) {
 	p.logger = logger
 }
 
+// SetID sets the ICMP identifier
```
SuperQ, line 391:
```suggestion
// SetID sets the ICMP identifier.
```

ping.go
```diff
@@ -385,6 +388,16 @@ func (p *Pinger) SetLogger(logger Logger) {
...
+func (p *Pinger) SetID(id int) {
+	p.id = id
+}
+
+// ID returns the ICMP identifier
```
SuperQ, line 396:
```suggestion
// ID returns the ICMP identifier.
```

natesales:
Thanks for the review! I've made the proposed changes and squashed into one commit.

SuperQ (approved):
LGTM

CHTJonas (approved):
LGTM

PR #196, opened 2022-01-03
#195 count packets of which  writes fail still as sent packets + exit run loop in case count is reached

### LINKED ISSUE #195: ping endlessly running in case the network interface is lost while the pinger is not yet finished
My System:
- Ubuntu Linux
- connected via Wi-Fi

Steps to reproduce:
- Take care that your PC is connected to the internet
- Start a ping run
  - with a count > 1 (I did choose 3)
  - Set an interval so that you have enough time to kill your interface (I did choose 1second which results in 3s time)
- now turn off your network interface (I did turn of Wi-Fi)

In the console the following can be seen endlessly
```
...
2021/12/27 10:35:42 FATAL: sending packet: write udp 0.0.0.0:77->8.8.8.8:0: sendto: network is unreachable
2021/12/27 10:35:43 FATAL: sending packet: write udp 0.0.0.0:77->8.8.8.8:0: sendto: network is unreachable
2021/12/27 10:35:44 FATAL: sending packet: write udp 0.0.0.0:77->8.8.8.8:0: sendto: network is unreachable
2021/12/27 10:35:45 FATAL: sending packet: write udp 0.0.0.0:77->8.8.8.8:0: sendto: network is unreachable
...
```

In the code I saw a "FIXME" where it hangs (line 485)
```
		case r := <-recvCh:
			err := p.processPacket(r)
			if err != nil {
				// FIXME: this logs as FATAL but continues
				logger.Fatalf("processing received packet: %s", err)
			}
```

turtletramp:
Based on my comparison about how ping in Ubuntu works here is a potential fix which counts packets that cannot be written still as sent packets. The result is equal statistics to standard ping command.

potentially fixes #195 

This approach fails one of the already existing tests which I also modified.

Any suggestions on how to fix this differently is welcome.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -34,7 +34,7 @@
 //	pinger.OnFinish = func(stats *ping.Statistics) {
 //		fmt.Printf("\n--- %s ping statistics ---\n", stats.Addr)
 //		fmt.Printf("%d packets transmitted, %d packets received, %v%% packet loss\n",
-//			stats.PacketsSent, stats.PacketsRecv, stats.PacketLoss)
+//			stats.CountExecuted, stats.PacketsRecv, stats.PacketLoss)
 //		fmt.Printf("round-trip min/avg/max/stddev = %v/%v/%v/%v\n",
 //			stats.MinRtt, stats.AvgRtt, stats.MaxRtt, stats.StdDevRtt)
 //	}
@@ -134,7 +134,10 @@ type Pinger struct {
 	// Debug runs in debug mode
 	Debug bool
 
-	// Number of packets sent
+	// CountExecuted is the number of ping executions that were started. If PacketsRecv == CountExecuted all pings were successful.
+	CountExecuted int
+
+	// Number of packets sent; is < CountExecuted in case errors happened during send procedure
 	PacketsSent int
 
 	// Number of packets received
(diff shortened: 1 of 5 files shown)


turtletramp:
I checked the circleci info and I do not see how I could fix these; looks like some updates happened and as a result some hashes do not match anymore.

SuperQ:
Yea, looks like something broke with goreleaser.  I will see if I can fix
it.

ping.go
```diff
@@ -734,14 +736,17 @@ func (p *Pinger) sendICMP(conn packetConn) error {
...
 			if neterr, ok := err.(*net.OpError); ok {
 				if neterr.Err == syscall.ENOBUFS {
 					continue
 				}
 			}
+			p.PacketsSent++
 			return err
 		}
+
 		handler := p.OnSend
```
SuperQ, line 750: I wonder if we also need to deal with the `p.OnSend` as well.

turtletramp: An interesting point...

In my change I only wanted to fix the problem with the endlessly running ping that will not stop anymore.
My change now includes also the tried sends in the count of PacketsSent and not only the successful ones as before. 
And as I do not know what the original intention of the p.OnSend was and what it is already used for, I cannot provide any meaningful opinion to that.

To keep it simpler I would see the topic "should p.OnSend be only executed for successful sends" independent of the goal of my fix and worth to create a new, separate issue to discuss it ;)

SuperQ: Looking this over again, I think this should be a separate counter, rather than increasing sent, since we didn't actually send anything

turtletramp: I fully agree; I was struggling myself to go the simple path with just increasing PacketsSent because I feared the code would get more complicated. But I will give it a try and will check how it could look like.

SuperQ:

PR #197, opened 2022-01-06
Delete redundant pinger.Stop() call

chenlujjj:
Since it's already in 
```
	g.Go(func() error {
		defer p.Stop()
		return p.runLoop(conn, recv)
	})
```


SuperQ (approved):
LGTM

CHTJonas (approved):
LGTM 👍🏻

Issue #198, opened 2022-01-10
ping result is always same for different addresses 


SHA65536:
Hello! I tried making a program that will ping a list of IPs and gather results.
```go
func scrapeAll() {
	pinger := ping.New("0.0.0.0")
	// Tried this too
	// pinger, _ := ping.NewPinger("0.0.0.0")
	pinger.Count = 5
	pinger.Size = 24
	pinger.Interval = time.Second
	pinger.Timeout = time.Second * 3
	pinger.TTL = 64
	pinger.SetPrivileged(true)
	pinger.OnFinish = handlePing
	uids, hosts := getEnabledDevices()
	for i := range uids {
		pinger.SetAddr(hosts[i])
		err := pinger.Run()
		if err != nil {
			consoleLog.Printf("Failed to ping target host %s, %v",hosts[i], err)
			continue
		}
	}
}
func handlePing(stats *ping.Statistics) {
	consoleLog.Printf("%v - %v", stats.Addr, stats.AvgRtt)
}
```
When running I get the result:
```
2022/01/10 11:20:10 127.0.0.1 - 197.967µs
2022/01/10 11:20:10 8.8.8.8 - 197.967µs
2022/01/10 11:20:10 192.168.5.7 - 197.967µs 
```
Tested multiple times so it's not a coincidence.
Maybe is there some internal variable in the pinger I need to reset before each ping?
Cheers.

go version go1.17.6 windows/amd64

SHA65536:
It seems that in runLoop you check packets ` p.PacketsRecv >= p.Count ` and ` p.PacketsSent >= p.Count `
So when changing the address using SetAddr the p.PacketsRecv and p.PacketsSent and other stats are not getting reset.

CHTJonas:
Please don't do that. `ping.Pinger` instances aren't designed to be used to ping multiple targets in that way.

Initialise a new `Pinger` for each host inside the `for` loop.

PR #199, opened 2022-01-10
Reuse Pinger feature

SHA65536:
Hey I recently needed to ping a list of addresses and I think that instead of needing to create a new pinger each time, it would be easier to just use the same pinger.
As I understood it's not currently possible since the internal state of the pinger does not change when using SetAddr.

Problem and replication described in #198 
I saw someone in the past tried to solve this issue, not sure why the pull request didn't get merged: #117 

659841dbad04df285d05946f9439118f1fcf8729 + 7613ef775df355ab0163a96a66aff5184ed8c270
I added a Reset function to be able to manually reset the internal state of the pinger, specifically packets received and sent, and minimum and maximum Rtt.

2741d0bb576648eee3ec9568d9ce49345f95834b
Added the Reset function to SetAddr so the statistics won't get mixed after changing the address.
I don't think there's a use case for SetAddr unless you also want to re-use the pinger. 

I'm not well versed in github so I apologize if something about this PR is wrong!
Cheers! 😄

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -348,6 +348,7 @@ func (p *Pinger) Resolve() error {
 func (p *Pinger) SetAddr(addr string) error {
 	oldAddr := p.addr
 	p.addr = addr
+	p.Reset()
 	err := p.Resolve()
 	if err != nil {
 		p.addr = oldAddr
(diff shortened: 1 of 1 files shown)


SuperQ:
Sorry, but this is not what you want to do. You want to re-create the pinger each time with `ping.New()`. This is fine and normal to do in Go. Just move the `ping.New(...)` inside your for loop.

PR #205, opened 2022-02-25
Report whether or not the TTL value was read

mem:
In IPv6 a hop limit of 0 is valid, so the TTL the code is reporting is
actually a valid value in that case.

Initialize ttl to -1, so that in the case of not getting a TTL value
from the control message, we can let the rest of the code whether the
reported value was actually 0 or whether it couldn't be read.


SuperQ:
It looks like we need to fix the goreleaser setup.

mem:
> It looks like we need to fix the goreleaser setup.

https://github.com/go-ping/ping/pull/210

CHTJonas (approved):

SuperQ (approved):

PR #210, opened 2022-04-26
Goreleaser can no longer be installed using goreleaser.sh

mem:
This installation method was deprecated because the associated tool has
been deprecated.

Download "by hand" and run checks before trying to use binary.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 executors:
   golang:
     docker:
-    - image: cimg/go:1.15
+    - image: cimg/go:1.17
 
 version: 2.1
 jobs:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 executors:
   golang:
     docker:
-    - image: cimg/go:1.15
```
mem, line 5: darwin/arm64 fails to build with this version

.circleci/config.yml
```diff
@@ -14,8 +14,12 @@ jobs:
...
       - run: golangci-lint run
       - run: make
-      - run: curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | BINDIR=/home/circleci/.local/bin sh
-      - run: goreleaser release --skip-publish --snapshot
+      - run: cd /tmp && curl -L --remote-name https://github.com/goreleaser/goreleaser/releases/download/v1.8.3/goreleaser_Linux_x86_64.tar.gz
+      - run: cd /tmp && curl -L --remote-name https://github.com/goreleaser/goreleaser/releases/download/v1.8.3/checksums.txt
+      - run: cd /tmp && sha256sum --ignore-missing -c checksums.txt
+      - run: mkdir -p /home/circleci/.local/bin
+      - run: tar -C /home/circleci/.local/bin -xf /tmp/goreleaser_Linux_x86_64.tar.gz goreleaser
+      - run: goreleaser release --skip-publish --snapshot --debug
```
mem, line 22: sorry about the hacky method...

utils_linux.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux
```
mem, line 1: this is because of the newer Go version

.circleci/config.yml
SuperQ: It's IMO better than `curl | bash`. :joy_cat:

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@
 executors:
   golang:
     docker:
-    - image: cimg/go:1.15
+    - image: cimg/go:1.17
```
SuperQ, line 5: Want to bump all the way to 1.18?

SuperQ (approved):
LGTM, thanks for fixing this.

CHTJonas (approved):
Nice one - thanks! 🚀

.circleci/config.yml
mem: I saw an issue in the goreleaser repo about not working with 1.18 yet, and I did not want to make the can of worms even bigger :-P

.circleci/config.yml
mem: sure! but I did not want to pack this into a script with less repetition, etc, just because I didn't want to deal with figuring out the CircleCI environment on top of trying to fix this.

I'll follow up with something that moves this to a script that's called from here

PR #214, opened 2022-06-27
Remove uses of Fatalf which causes the application to quit if a custom logger is provided such as Uber zap

stanimirivanovde:
Uber zap's logger does `os.Exit()` if `Fatal()` or `Fatalf()` is called. Since go-ping supports custom loggers and in my case I pass in Uber zap's logger this causes my entire application to exit. Removing `Fatal()` calls is safer. If we need to ensure this error is handled up the call chain we can `panic()` which is better since the code can recover from a panic. The code cannot recover from an `os.Exit()` call.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -481,8 +481,8 @@ func (p *Pinger) runLoop(
 		case r := <-recvCh:
 			err := p.processPacket(r)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


SuperQ:

Issue #216, opened 2022-07-26
In "mac", the call "pinger.Run()" struck for ever


declum:
Code:
>
> pinger, err := goping.NewPinger("www.google.com")
> 	if err != nil {
> 		panic(err)
> 	}
> 	pinger.Count = 3
> 	pinger.Run()                 // blocks until finished
> 	stats := pinger.Statistics() // get send/receive/rtt stats
> 	log.Println(stats)
>

this part of the loop is getting executed. But the loop doesnt end. Can we making `continue` as `return nil` works. Looks like interval.Stop() doesnt end the tick.
> 
>			if p.Count > 0 && p.PacketsSent >= p.Count {
>				interval.Stop()
>				continue
>			}
>

$ go version
go version go1.18.4 darwin/amd64

$ uname -a
Darwin XXXXX 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64 x86_64

Issue #180, opened 2021-09-11
Non-privileged scenarios do not work


HuKeping:
After commit 62f79f1f4f36fc821eebb131548270d4d8b68653 upstream, the go-ping program blocks the received ICMP packets because the received ICMP identifier is different from the senter's echo identifier

We think it may caused by  the Linux kernel automatically assigns an identifier as the local receive port number in non-privileged cases.

HuKeping:
ping @mem

CHTJonas:
I'm not experiencing this issue.

```
chtj2@pip:/tmp/chtj2$ uname -a
Linux pip 5.4.0-74-generic #83-Ubuntu SMP Sat May 8 02:35:39 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

chtj2@pip:/tmp/chtj2$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal

chtj2@pip:/tmp/chtj2$ go version
go version go1.15.2 linux/amd64

chtj2@pip:/tmp/chtj2$ git clone https://github.com/go-ping/ping.git
Cloning into 'ping'...
remote: Enumerating objects: 290, done.
remote: Counting objects: 100% (23/23), done.
remote: Compressing objects: 100% (17/17), done.
remote: Total 290 (delta 8), reused 12 (delta 4), pack-reused 267
Receiving objects: 100% (290/290), 1.28 MiB | 2.00 MiB/s, done.
Resolving deltas: 100% (144/144), done.

chtj2@pip:/tmp/chtj2$ cd ping/

chtj2@pip:/tmp/chtj2/ping$ go run cmd/ping/ping.go google.com
go: downloading golang.org/x/net v0.0.0-20210316092652-d523dce5a7f4
go: downloading golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
go: downloading github.com/google/uuid v1.2.0
go: downloading golang.org/x/sys v0.0.0-20210315160823-c6e025ad8005
PING google.com (142.250.187.238):
32 bytes from 142.250.187.238: icmp_seq=0 time=6.165858ms ttl=116
32 bytes from 142.250.187.238: icmp_seq=1 time=6.051684ms ttl=116
32 bytes from 142.250.187.238: icmp_seq=2 time=6.041586ms ttl=116
[...]

PR #218, opened 2022-08-27
Fix non-privileged (UDP) ping response parsing for OSes not supporting STRIPHDR sock option

emakeev:
It could be related to the issues #180 & #216. I noticed that on Mac OS 12.5, MacBook Pro M1 non-privileged pings do not work on remote IPs (pinging local loopback interfaces works fine). 
While I can see correct  ICMP requests & responses with tcpdump, ping discards responses because they include  unexpected IP headers.

It looks like STRIPHDR socket option is not working as expected on some darwin setups: https://go.googlesource.com/net/+/master/icmp/listen_posix.go#75.

The workaround fixes the issue on Mac & doesn't break working loopback pings, but it wasn't tested on Windows or linux.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /ping
 /dist
+.idea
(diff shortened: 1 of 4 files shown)


SuperQ:

PR #222, opened 2022-09-29
Flexible ping interval option (allows to send next ping packet as soon as the previous packet echo answer is received)

emakeev:
Add a pinger flag to allow sending consecutive ping packets as soon as preceding packet's echo response was successfully received & parsed.


SuperQ:

Issue #221, opened 2022-09-22
Is this project abandoned?


reimda:
I'm a maintainer of the telegraf project. We have received a [request](https://github.com/influxdata/telegraf/pull/11836) to migrate away from go-ping and use a fork, [pro-bing](https://github.com/prometheus-community/pro-bing).

It's been four months since there's been a commit here and there are a few PRs that need maintainer attention. @CHTJonas @sparrc could you respond and share your plans for the org and repo? Should projects like telegraf migrate to a fork if they need module dependency updates, bugfixes, or new features?

SuperQ:
As one of the 3 `go-ping/ping` maintainers, I vote abandoned.

CHTJonas:
As one of the other maintainers, I second @SuperQ.

CHTJonas:
Suggested fork is https://github.com/prometheus-community/pro-bing

PR #224, opened 2022-11-26
Add packet timeout option and make latency more accurate

flizaga:
Two major improvements :
- Add packet timeout option
- Improve ping rtt time accuracy on windows (to be able to go under millisecond)

Small improvement : 
- Use program side time object instead of storing datetime on the packet body to get ping packet rtt. It avoids bad timing while clock is changing (for DST or on NTP sync, for example) ref : https://github.com/golang/go/issues/12914

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -88,14 +88,15 @@ var (
 func New(addr string) *Pinger {
 	r := rand.New(rand.NewSource(getSeed()))
 	firstUUID := uuid.New()
-	var firstSequence = map[uuid.UUID]map[int]struct{}{}
-	firstSequence[firstUUID] = make(map[int]struct{})
+	var firstSequence = map[uuid.UUID]map[int]uint64{}
+	firstSequence[firstUUID] = make(map[int]uint64)
 	return &Pinger{
 		Count:      -1,
 		Interval:   time.Second,
 		RecordRtts: true,
 		Size:       timeSliceLength + trackerLength,
 		Timeout:    time.Duration(math.MaxInt64),
+		PktTimeout: time.Duration(math.MaxInt64),
 
 		addr:              addr,
 		done:              make(chan interface{}),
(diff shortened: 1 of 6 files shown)


fkr:
@flizaga I came across the timeout a while ago and also wanted to add a default timeout - however if I understand correctly you chose 15 seconds default timeout. I would've gone for 10 (which is the default in the linux and BSD implementations of ping). (side note: windows has 4 seconds default timeout).

ping.go
```diff
@@ -478,6 +487,16 @@ func (p *Pinger) runLoop(
 		case <-timeout.C:
 			return nil
 
+		case <-ptimeout.C:
```
SuperQ, line 490: This approach has been proposed before. Similar to #176 and #202.

The problem is this only checks for packet timeouts every time the ticker runs, which means the per-packet timeouts are not going to be accurate.

For example, if you have an interval of 1s timeout of 15s, the second packet won't actually timeout until the second interval at 30s. This means it will have timed out at 29 seconds.

I think in order to do this correctly we need to reset the PktTimeout timer each time the oldest packet is received. This way we inch the timeout forward to the next un-received packet.

SuperQ:
FYI: This project is abandoned as we are unable to maintain it due to lack of maintainers with admin rights on the repo. See #221

flizaga:
Well, indeed, regarding the default timeout, I should have used a system default one, as it's possible to set it up to 10 for linux and BSD and 4 to windows.
If I had think a bit much more I should have set the Timeout ticker every one seconds, as I don't think people would need a timeout under a second, and usually everyone will setup a integer number of seconds for the packet timeout.

Regarding the other topics : better accuracy on ping RTT both on linux/bsd and windows (and finally a ping that goes under the ms for the RTT (default Windows ping is not able to do it), what do you think ?

P.S. : it's the first time I code in Go, so I hope my code is respecting the coding standards.

SuperQ:

PR #236, opened 2024-11-22
Mark module as deprecated

Hipska:
In addition to #226, also mark the Go package as deprecated.

### DIFF
--- a/ping.go
+++ b/ping.go
@@ -1,4 +1,5 @@
 // Deprecated: This package is no longer maintained.
+// Package ping is a simple but powerful ICMP echo (ping) library.
 //
 // Here is a very simple example that sends and receives three packets:
 //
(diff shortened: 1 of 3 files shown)


ping.go
```diff
@@ -1,3 +1,4 @@
+// Deprecated: This package is no longer maintained.
```
SuperQ, line 1: I think this needs to be next to `package ping`.

ping.go
```diff
@@ -1,3 +1,4 @@
...
 // Package ping is a simple but powerful ICMP echo (ping) library.
 //
```
SuperQ, line 3:
```suggestion
// Package ping is a simple but powerful ICMP echo (ping) library.
//
// Deprecated: This package is no longer maintained.
```

SuperQ (approved):

ping.go
Hipska: I added it just above `package ping` as your previous comment. What to do?

SuperQ: Either works according to the [Go docs](https://go.dev/wiki/Deprecated#examples). Up to you.

SuperQ (approved):

CHTJonas (approved):

PR #238, opened 2025-11-26
Fix slice backing array sharing in Statistics()

pawciobiel:


SuperQ:
Can you explain more or add some tests that show what is being fixed?

